PR #5, opened 2020-09-22
Use finalizer to cleanup DB on CR delete

stuggi:
Introduces a finalizer to run a DB delete job on CR delete to be
able to cleanup also DB from other operators.

### DIFF
--- a/controllers/mariadbdatabase_controller.go
+++ b/controllers/mariadbdatabase_controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -27,6 +28,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	util "github.com/openstack-k8s-operators/lib-common/pkg/util"
 	databasev1beta1 "github.com/openstack-k8s-operators/mariadb-operator/api/v1beta1"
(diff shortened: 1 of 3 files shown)


dprince (approved):

controllers/mariadbdatabase_controller.go
```diff
@@ -76,10 +75,59 @@ func (r *MariaDBDatabaseReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
 	if err != nil {
 		if k8s_errors.IsNotFound(err) {
 			r.Log.Info("No DB found for label 'dbName'.")
+			return ctrl.Result{}, nil
 		}
 		return ctrl.Result{RequeueAfter: time.Second * 10}, err
 	}
 
+	finalizerName := "mariadb-" + instance.Name
```
mdbooth, line 83: Why not make it a constant?

controllers/mariadbdatabase_controller.go
```diff
@@ -76,10 +75,59 @@ func (r *MariaDBDatabaseReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
...
+	// if deletion timestamp is set on the instance object, the CR got deleted
+	if instance.ObjectMeta.DeletionTimestamp.IsZero() {
+		// if it is a new instance, add the finalizer
+		if !controllerutil.ContainsFinalizer(instance, finalizerName) {
+			controllerutil.AddFinalizer(instance, finalizerName)
+			err = r.Client.Update(context.TODO(), instance)
+			if err != nil {
+				return ctrl.Result{}, err
+			}
+			r.Log.Info(fmt.Sprintf("Finalizer %s added to CR %s", finalizerName, instance.Name))
```
mdbooth, line 93: Incidentally, CreateOrUpdate is really good at this. You just define it in your create function, and it will be added automatically to all your existing objects when you update them.

controllers/mariadbdatabase_controller.go
stuggi: yes correct, could be a const. I wanted to have the instance name in the secret, but it is actually not required.

controllers/mariadbdatabase_controller.go
```diff
@@ -76,10 +75,59 @@ func (r *MariaDBDatabaseReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
...
+		if err != nil {
+			return ctrl.Result{}, err
+		} else if requeue {
+			r.Log.Info("Waiting on DB delete")
+			return ctrl.Result{RequeueAfter: time.Second * 5}, err
+		}
+
+		// delete the job
+		requeue, err = util.DeleteJob(job, r.Kclient, r.Log)
+		if err != nil {
```
bogdando, line 117: should a log message be logged as well?

PR #6, opened 2020-09-22
Allow underscore in db names like nova_api, nova_cell0

stuggi:
This is not a full polished patch, just removes the invalid
underscore character from the metadata name.

### DIFF
--- a/pkg/database.go
+++ b/pkg/database.go
@@ -1,6 +1,8 @@
 package mariadb
 
 import (
+	"strings"
+
 	util "github.com/openstack-k8s-operators/lib-common/pkg/util"
 	databasev1beta1 "github.com/openstack-k8s-operators/mariadb-operator/api/v1beta1"
 	batchv1 "k8s.io/api/batch/v1"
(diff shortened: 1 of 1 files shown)


pkg/database.go
```diff
@@ -22,7 +24,7 @@ func DbDatabaseJob(database *databasev1beta1.MariaDBDatabase, databaseHostName s
 	}
 	job := &batchv1.Job{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      database.Spec.Name + "-database-sync",
+			Name:      strings.Replace(database.Spec.Name, "_", "", -1) + "-database-sync",
```
dprince, line 27: may add a comment as to why we do the replace? Also, this means if someone submits a CR named nova_api and novaapi they'd both end up the same right?

stuggi: commend added in updated commit. yes if the dbname has an underscore it gets removed in the metadata name, so nova_api and novaapi result in the same.

PR #8, opened 2020-09-24
Comment out controller tests for now

dprince:


mpryc:
@dprince Thanks for this. I have combined this one (with additional commenting out of some imports to make vet happy) together with other small lint nits and seems like travis is passing now:
  https://github.com/openstack-k8s-operators/mariadb-operator/pull/7

If possible we should abandon this change and merge https://github.com/openstack-k8s-operators/mariadb-operator/pull/7

PR #12, opened 2020-10-01
Fix race with MariaDB/MariaDBDatabase

dprince:
This resolves a race introduced in 5d6e4248 where MariaDBDatabase
objects might not get created unless the MariaDB already exists.
The issue was that reconcile was returning a nil result (thus
signalling completion) when it was supposed to be blocking
on the MariaDB to be created.

### DIFF
--- a/controllers/mariadbdatabase_controller.go
+++ b/controllers/mariadbdatabase_controller.go
@@ -75,11 +75,10 @@ func (r *MariaDBDatabaseReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
 	objectKey, err := client.ObjectKeyFromObject(db)
 	err = r.Client.Get(context.TODO(), objectKey, db)
 	if err != nil {
-		if k8s_errors.IsNotFound(err) {
-			r.Log.Info("No DB found for label 'dbName'.")
-			return ctrl.Result{}, nil
+		if !k8s_errors.IsNotFound(err) {
+			return ctrl.Result{}, err
 		}
-		return ctrl.Result{RequeueAfter: time.Second * 10}, err
+		return ctrl.Result{RequeueAfter: time.Second * 10}, nil
 	}
 
 	finalizerName := "mariadb-" + instance.Name
--- a/controllers/suite_test.go
+++ b/controllers/suite_test.go
@@ -29,7 +29,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/envtest/printer"
 	// logf "sigs.k8s.io/controller-runtime/pkg/log"
 	// "sigs.k8s.io/controller-runtime/pkg/log/zap"
-
 	// databasev1beta1 "github.com/openstack-k8s-operators/mariadb-operator/api/v1beta1"
 	// +kubebuilder:scaffold:imports
 )
@@ -48,6 +47,7 @@ func TestAPIs(t *testing.T) {
 		"Controller Suite",
 		[]Reporter{printer.NewlineReporter{}})
 }
+
 /*
 var _ = BeforeSuite(func(done Done) {
 	logf.SetLogger(zap.LoggerTo(GinkgoWriter, true))


controllers/mariadbdatabase_controller.go
```diff
@@ -75,11 +75,10 @@ func (r *MariaDBDatabaseReconciler) Reconcile(req ctrl.Request) (ctrl.Result, er
 	objectKey, err := client.ObjectKeyFromObject(db)
 	err = r.Client.Get(context.TODO(), objectKey, db)
 	if err != nil {
-		if k8s_errors.IsNotFound(err) {
-			r.Log.Info("No DB found for label 'dbName'.")
-			return ctrl.Result{}, nil
+		if !k8s_errors.IsNotFound(err) {
+			return ctrl.Result{}, err
```
mdbooth, line 79: This isn't really an error in this context. This is an expected situation which we can handle. We don't need a stack trace. I believe the best action here would be to requeue:

return ctrl.Result{Requeue: true}, nil

Possibly use RequeueAfter instead of Requeue if you have a good idea how long it might take.

The original log message could also stay, btw.

mdbooth (commented):
I've hit a ton of problems like this in ovn-central. This one's a code bug, but there are other situations where you see:

```
if ! Check on thing A {
  return
}
...
fix thing A
```

So you look and you reverse the order, but that just has the same problem for some other object. I think I read in some best practise document that you change one thing and then return, so that's how I started out. However, after a bunch of headaches, and realising that sometimes there's no order that can never result in starvation of something, I came to the decision:

Always do all the things. Stop for errors, nothing else.

So with this I'd structure this as:

```
db, err := doDBThings()
stop on error

err := finalizeAllTheThings()
stop on error

doDBThings():
  if Get(db) == nil return nil, nil
  doMoreThings()
```

At the very least, it's weird we get the db, run the finalizer, then use the db.

controllers/mariadbdatabase_controller.go
dprince: Two things. We always requeue due to the line below this. Second, we don't want the log messages as it causes tons of Logs when you deploy the mariadb-operator *if* there is no MariaDB database deployed. I've tested the code I have here well and what I have now works in practice, and should provide the logs we need if there is an error (aside from the expected NotFound API error when you haven't created the MariaDB on initial mariadb-operator install).

dprince: The RequeueAfter: time.Second IMO is a better implementation. May as well wait a few seconds on the database to initialize as it takes at least that long...

dprince:
> I've hit a ton of problems like this in ovn-central. This one's a code bug, but there are other situations where you see:
> 
> ```
> if ! Check on thing A {
>   return
> }
> ...
> fix thing A
> ```
> 
> So you look and you reverse the order, but that just has the same problem for some other object. I think I read in some best practise document that you change one thing and then return, so that's how I started out. However, after a bunch of headaches, and realising that sometimes there's no order that can never result in starvation of something, I came to the decision:
> 
> Always do all the things. Stop for errors, nothing else.
> 
> So with this I'd structure this as:
> 
> ```
> db, err := doDBThings()
> stop on error
> 
> err := finalizeAllTheThings()
> stop on error
> 
> doDBThings():
>   if Get(db) == nil return nil, nil
>   doMoreThings()
> ```
> 
> At the very least, it's weird we get the db, run the finalizer, then use the db.

If we want to re-work the finalizer flow I'm up for it. But I think that is a separate patch conversation right? What I'm doing here is fixing something that is functionally broken (a race condition). If you read the commit it was actually a side effect that probably shouldn't have gone into the finalizer patch though.

This has broken my environment badly for over a week now. Just really want to get the immediate issue fixed for everyone.

controllers/mariadbdatabase_controller.go
mdbooth: Wait, I didn't notice you flipped the logic!

Yeah, this is good.

mdbooth (approved):
The code structure thing is more food for thought. I have plenty of it :)

stuggi (approved):

PR #20, opened 2022-05-27
set user in container spec

SeanMooney:
This change add runAsUser to the
security context to avoid the following error:

container has runAsNonRoot and image has non-numeric user (nonroot),
cannot verify user is non-root

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,7 +33,7 @@ RUN cp -r templates ${DEST_ROOT}/templates
 FROM $OPERATOR_BASE_IMAGE
 
 ARG DEST_ROOT=/dest-root
-ARG USER_ID=nonroot:nonroot
+ARG USER_ID=65532
 
 ARG IMAGE_COMPONENT="mariadb-operator-container"
 ARG IMAGE_NAME="mariadb-operator"
@@ -56,7 +56,7 @@ LABEL com.redhat.component="${IMAGE_COMPONENT}" \
       io.openshift.tags="${IMAGE_TAGS}"
 ### DO NOT EDIT LINES ABOVE
 
-ENV USER_UID=1001 \
+ENV USER_UID=$USER_ID \
     OPERATOR_TEMPLATES=/usr/share/mariadb-operator/templates/ \
     WATCH_NAMESPACE=openstack,openshift-machine-api,openshift-sriov-network-operator
 


dprince:
@SeanMooney just curious what container image you are using to hit the error you mention?

SeanMooney:
quay.io/tripleotraincentos8/centos-binary-mariadb@sha256:62813d789acb69231965769ba487ab7fe4e522b97be601be013cfb2a06e349f8

for mariadb https://termbin.com/pewp

deployed by https://termbin.com/bq5au

i worked around it locally by editing the deployment deployment.apps/mariadb-operator-controller-manager  directly and added  runAsUser: 1000

dprince:
How are you deploying the operator to hit this? Just curious why nobody else is seeing this... (it works for me)

SeanMooney:
it was deployed using the install_yaml makefiles using microsfhist to provide the OpenShift installation.

I automated the installation of microshfit 
https://github.com/SeanMooney/ansible_role_devstack/commit/647e80d02313ecd0d13404ab041f1e18d300358f
and old using operator sdk 1.21
https://github.com/SeanMooney/ansible_role_devstack/commit/1e292aec81172eafacc0fe93fc66fb789409d0e6

I then manually installed kustomise and https://github.com/rancher/local-path-provisioner to provide storage. I modified the latter to have the same storage class name that install_yamls expects. I have used it in the pass when I deploy Zuul/nodepool/gerriita and caddy to run a third party ci hosted by k8s from home it work pretty well and is much simpler to use the the crc_stroage target that install_yamml currently uses.

I still need to automate this with ansibale but basically I just did 
molecule create -s microshift
molecule converge -s micorshift
molecule login -s micorshift -h microshift
sudo su stack
curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | bash
sudo mv kustomize /usr/bin
wget https://raw.githubusercontent.com/rancher/local-path-provisioner/v0.0.22/deploy/local-path-storage.yaml
sed /name: local-path/name: local-storage/ -i local-path-storage.yaml
sudo mkdir -m 777 /opt/local-path-provisioner 
sudo setenforce permissive

# not sure I need to set selinux to permissive by the way I created the local path provision before I created the directory and it got permission denied issues until I
deleted it and recreated it, setenforce permissive didn't actually help fix it by its self until I deleted and recreated it so I think I can remove that when I automated it.

kubctl apply -f local-path-storage.yaml
git clone https://github.com/openstack-k8s-operators/install_yamls
cd install_yamls
make marinadb

at this point the manager pod was in error "CreateContainerConfigError"

I then fixed that by doing "oc edit deployment.apps/mariadb-operator-controller-manager"
setting replicas to 0 and updating the runAsUser line the setting replicas to 1

then I ran "watch oc get all" until it fixed iself 
and could then run.
make marinadb_deploy


This isn't the exact set of command I used but is close enough and it what I'm basically planning to automate via ansible via a new roles.

long term I'm probably going to stop using the install_yamls repo as I hate make and implement the same logic with an ansibel role or several but my goal with the hack and hussel day we had on Friday was to try and replicate the demo of deploying keystone.

if I had time I was then going to use my devstack ansibel roles to deploy a controller without keystone using keystone form OpenShift and a compute node. the idea being as we modify more services and or complete the standalone ansible roles in ooo I can swap out services one by one until we have just a  podified  controlplane with ansibel external compute. in the interim dev stack and bridge the gap installing any service the others can't.

SeanMooney:
this is the check that is failing by the way that cause the error

https://github.com/kubernetes/kubernetes/blob/5648200571889140ad246feb82c8f80a5946f167/pkg/kubelet/kuberuntime/security_context.go#L88

which is called form here
https://github.com/kubernetes/kubernetes/blob/5648200571889140ad246feb82c8f80a5946f167/pkg/kubelet/kuberuntime/kuberuntime_container.go#L196

as I noted on my other pr to the keystone operator the other way to fix this is to fix the regression into ducted when you added the GitHub workflow.

prior to that commit the containers had numeric user id

https://github.com/openstack-k8s-operators/keystone-operator/pull/55#issuecomment-1140133245

actually your reel docker file actually is not broken I would guess that is why you are not seeing it

the normal one has USER $USER_ID
https://github.com/openstack-k8s-operators/mariadb-operator/blob/master/Dockerfile#L71
$USER_ID is nonroot:nonroot
https://github.com/openstack-k8s-operators/mariadb-operator/blob/119f8d825a716bf7d218c07e8d4562f8d8bf20b4/Dockerfile#L36

the rhel version has
USER ${USER_UID}
https://github.com/openstack-k8s-operators/mariadb-operator/blob/master/Dockerfile.rhel#L64

where $USER_UID is 1001
https://github.com/openstack-k8s-operators/mariadb-operator/blob/119f8d825a716bf7d218c07e8d4562f8d8bf20b4/Dockerfile.rhel#L45

so a better fix is probably to fix the docker file but I suspect others are not seeing this because they are using the RHEL images not the default one.
if you

SeanMooney:
im going to work on a slight update to this but based on 
https://github.com/GoogleContainerTools/distroless/blob/main/examples/nonroot/BUILD#L16=
i think the nonroot user is actually 1002 by default
although we were i think expecting to use 1001

I'm going to try and update the docker file and make things consistent

SeanMooney:
ah found it its 65532

https://github.com/GoogleContainerTools/distroless/blob/main/base/base.bzl#L8=

SeanMooney:
dining in to this more it looks like Microshift is currently based on 4.8
th 4.10 and 4.11 rebases are currently in flight but that is likely the reason for the delta with crc.

I deployed with arc today using 4.10 and confirm the operator works as intended without this change.

the check that is failing with 4.8 was first introduced by 
https://github.com/kubernetes/kubernetes/commit/445393fdcefa6d0354b7ce32a2304a7765fbd305
that check was later moved to a different location by 
https://github.com/kubernetes/kubernetes/commit/ccde63b9c16241780f42d25550c200aa38e53989#
when fixing the support for windows

we can see in the test code that setting un as root with a non numeric user name is expected to fail when using the pkg/kubelet/kuberuntime/security_context_others.go

https://github.com/kubernetes/kubernetes/commit/ccde63b9c16241780f42d25550c200aa38e53989#diff-e7192232501aa61e3b33aaf4c5ebc591326201275ffd4426eb72a0e7de3622e5R109-R115

the code on master still asserts it only supports numeric users

https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/kuberuntime/kuberuntime_container.go#L299-L314

its not clear to me at this time why crc does not have the same error but I can retry the deployment when micors hift updates to 4.10

config/manager/manager.yaml
```diff
@@ -34,6 +34,9 @@ spec:
         image: controller:latest
         name: manager
         securityContext:
+          # if we enable runAsNonRoot we must set the user
+          runAsUser: 65535
+          runAsGroup: 65535
```
dprince, line 39: I do not thing we want to hardcode these values here. The securityContext on the spec already containers runAsNonRoot which should be good enough.

The reason we probably don't want to hard code these here is due to the fact that there may be differences in the default nonroot UID between our upstream and downstream images (distroless, vs. RHEL ubi-minimal)

dprince:
@SeanMooney I had tried this for keystone. Have you tried just removing the USER_UID=1001 bit from the upstream Dockerfile like this: 

https://github.com/openstack-k8s-operators/keystone-operator/commit/3117815b387e74d93fcb80b4f8dfea6b64de3dc2

Index image built here: quay.io/dprince/keystone-operator-index:3117815b387e74d93fcb80b4f8dfea6b64de3dc2

NOTE: I tried this based on your keystone patch but you seem to be commenting more on things here. Might I suggest dropping one of your reviews until we agree on any needed updates?

SeanMooney:
yes w can focus on just one of the reviews, that fine by me.

no but that would not help the problem is that we must have a numeric userid if we have runAsNonRoot defined.

so removing the USER_UID=1001 won't work.

I just tried using the image you created and I got the same error.

https://termbin.com/korr

if I edit the deployment and add runAsUser: 65532 the check passes and the manager deploys.

https://termbin.com/r1qi

I believe we can drop the security context part if we just modify the upstream image to actually set numeric uuids in the docker file like we used too

https://github.com/openstack-k8s-operators/mariadb-operator/blame/f32af538099ba5f9f349f92724db2ca35705d931/Dockerfile#L25

setting the user id in the security context is only need because we currently set it to non root.

SeanMooney:
lets take the conversation to the keystone operator since that is what you prototyped.
i have build my pr and pushed it to the quay registry 

so

make keystone KEYSTONE_IMG=quay.io/smooney/keystone-operator-index:v0.0.1

now deploys for me with install.yaml

i can drop the security context changes and rebuild it localy and update quay

if that works i can update the pr if that works for you.

then if we are happy with that approch i can then update this mariadb pr too.

PR #21, opened 2022-06-15
Update to use latest lib-common (Jobs)

dprince:
This patch bumps the version of lib-common to the latest.

Initial changes to use shared Job helper functions.


stuggi (approved):
/lgtm

PR #23, opened 2022-07-01
Add TLS support

olliewalsh:
Add a TLS property to the CR to specify the the cert and CA secrets.
TLS/secretName and TLS/caSecretName seems to be the convention in other
operators.

Add secrets to rbac and fix missing patch support for pods. Fix Pod
CreateOrUpdate logic.

Trigger reconcile using a label watch as we do not own the TLS secrets.

Add volume mounts for the TLS secrets and kolla config.
Fix mariadb conf ownership, it was being silently ignored.


templates/mariadb/bin/galera.cnf
```diff
@@ -22,7 +22,10 @@ query_cache_limit = 1M
 query_cache_size = 16M
 skip-external-locking
 socket = /var/lib/mysql/mysql.sock
-ssl = false
+ssl-cert = /etc/pki/tls/certs/mariadb.crt
+ssl-cipher = !SSLv2:kEECDH:kRSA:kEDH:kPSK:+3DES:!aNULL:!eNULL:!MD5:!EXP:!RC4:!SEED:!IDEA:!DES:!SSLv3:!TLSv1
+ssl-key = /etc/pki/tls/private/mariadb.key
+ssl-ca = /etc/pki/tls/certs/mariadbca.crt
```
stuggi, line 28: does it still start without TLS? because we do not force TLS if I read the above correct.

olliewalsh:
Yes. The client decides to initiate the SSL handshake. If the SSL files
exist on the server then it will work. There is a conf option to force ssl
by default but it does not exist in this version of mariadb
https://mariadb.com/kb/en/server-system-variables/#require_secure_transport.

I believe we can require SSL for specific accounts though.

templates/mariadb/bin/galera.cnf
olliewalsh: Yes. The client decides to initiate the SSL handshake. If the SSL files
exist on the server then it will work. There is a conf option to force ssl
by default but it does not exist in this version of mariadb
https://mariadb.com/kb/en/server-system-variables/#require_secure_transport.

I believe we can require SSL for specific accounts though.

templates/mariadb/bin/galera.cnf
```diff
@@ -7,7 +7,7 @@ key_buffer_size = 16M
 
 [mysqld]
 basedir = /usr
-bind-address = 127.0.0.1
+bind-address = *
```
dprince, line 10: Do we need to listen on *?

olliewalsh: Yes, loopback would only be accessible within the pod. It's the default that we have been using all along since mysql user could not read this file.

controllers/mariadb_controller.go
```diff
@@ -142,25 +160,75 @@ func (r *MariaDBReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+			}
+			return ctrl.Result{}, err
+		}
+
+		tlsSecretHasLabel := tlsSecret.Labels[mariaDBReconcileLabel] == instance.Name
+		if !tlsSecretHasLabel {
+			if tlsSecret.Labels == nil {
+				tlsSecret.Labels = make(map[string]string)
+			}
+			tlsSecret.Labels[mariaDBReconcileLabel] = instance.Name
```
stuggi, line 181: could also use labels.Merge 
```
if value, ok := tlsSecret.Labels[mariaDBReconcileLabel]; !ok || value != instance.Name {
    tlsSecret.GetObjectMeta().SetLabels(labels.Merge(tlsSecret.GetObjectMeta().GetLabels(), tlsSecret.Labels[mariaDBReconcileLabel] = instance.Name)
    ... update secret
}
```

controllers/mariadb_controller.go
```diff
@@ -142,25 +160,75 @@ func (r *MariaDBReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+			err = r.Client.Update(ctx, tlsSecret)
+			if err != nil {
+				if k8s_errors.IsConflict(err) {
+					return ctrl.Result{Requeue: true}, err
+				}
+				if k8s_errors.IsNotFound(err) {
+					return ctrl.Result{Requeue: true}, err
+				}
```
stuggi, line 189: could do this in one if as the return ctrl is the same?

controllers/mariadb_controller.go
```diff
@@ -142,25 +160,75 @@ func (r *MariaDBReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+			}
+			return ctrl.Result{}, err
+		}
+
+		caSecretHasLabel := caSecret.Labels[mariaDBReconcileLabel] == instance.Name
+		if !caSecretHasLabel {
+			if caSecret.Labels == nil {
+				caSecret.Labels = make(map[string]string)
+			}
+			caSecret.Labels[mariaDBReconcileLabel] = instance.Name
```
stuggi, line 211: same as above, could use labels.Merge

templates/mariadb/bin/galera.cnf
olliewalsh: That was when using centos8/train images. If the target is centos9/wallaby then we should have require_secure_transport as it's mariadb 10.5.16.

templates/mariadb/bin/galera.cnf
olliewalsh: keystone-operator uses `Listen 0.0.0.0:5000`, which is basically the same but IPv4 only. In fact we should change keystone to `Listen 5000`.

stuggi: https://github.com/openstack-k8s-operators/keystone-operator/pull/86

olliewalsh:
Closing this initial POC PR. See https://github.com/openstack-k8s-operators/mariadb-operator/pull/119 instead

PR #27, opened 2022-07-12
Create separate module for API structs

dprince:


.github/workflows/build-mariadb-operator.yaml
```diff
@@ -91,7 +91,7 @@ jobs:
       uses: redhat-actions/openshift-tools-installer@v1
       with:
         source: github
-        operator-sdk: '1.18.1'
+        operator-sdk: '1.22.0'
```
stuggi, line 94: 1.22.1?

.github/workflows/build-mariadb-operator.yaml
```diff
@@ -82,7 +82,7 @@ jobs:
     - name: Install Go
       uses: actions/setup-go@v2
       with:
-        go-version: 1.17.x
+        go-version: 1.18.0
```
stuggi, line 85: 1.18.4?

.github/workflows/golangci-lint.yaml
```diff
@@ -32,7 +32,7 @@ jobs:
       - name: Install Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.17.8
+          go-version: 1.18.0
```
stuggi, line 35: current lastest 1.18.4 ?

.github/workflows/build-mariadb-operator.yaml
dprince: I'm not sure setup-go always supports the latest go version. But lets try it.

dprince: I think we might should just make all of these 1.18.x like what was there before.

stuggi (approved):
/lgtm

PR #28, opened 2022-07-14
Add OWNERS file

Sandeepyadav93:
Added OWNERS file required by PROW.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -1,0 +1,13 @@
+approvers:
+  - abays
+  - dprince
+  - mpryc
+  - olliewalsh
+  - stuggi
+
+reviewers:
+  - abays
+  - dprince
+  - mpryc
+  - olliewalsh
+  - stuggi


OWNERS
```diff
@@ -0,0 +1,13 @@
+approvers:
+  - abays
+  - dprince
+  - mpryc
```
stuggi, line 4: since Michal changed roles, I think we could remove him.

OWNERS
```diff
@@ -0,0 +1,13 @@
...
+  - olliewalsh
+  - stuggi
+
+reviewers:
+  - abays
+  - dprince
+  - mpryc
```
stuggi, line 11: same

mpryc: @stuggi agreed :-)

Sandeepyadav93: @stuggi @mpryc thanks for your review, I copied the OWNER file from osp-director-operator[1]. I was not aware Michal role changed. We need to update [1] as well - I will send a PR.

[1] https://github.com/openstack-k8s-operators/osp-director-operator/blob/master/OWNERS

PR #31, opened 2022-08-22
[Makefile] Update opm release page location

zerodayz:
I believe '$${ARCH} and $${OS}' is a typo and it should be '${ARCH} and ${OS}'

Considering that bash variable $$ resolving into a pid rather than arch and os.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -203,7 +203,7 @@ ifeq (,$(shell which opm 2>/dev/null))
 	set -e ;\
 	mkdir -p $(dir $(OPM)) ;\
 	OS=$(shell go env GOOS) && ARCH=$(shell go env GOARCH) && \
-	curl -sSLo $(OPM) https://github.com/operator-framework/operator-registry/releases/download/v1.19.1/$${OS}-$${ARCH}-opm ;\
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -203,7 +203,7 @@ ifeq (,$(shell which opm 2>/dev/null))
 	set -e ;\
 	mkdir -p $(dir $(OPM)) ;\
 	OS=$(shell go env GOOS) && ARCH=$(shell go env GOARCH) && \
-	curl -sSLo $(OPM) https://github.com/operator-framework/operator-registry/releases/download/v1.19.1/$${OS}-$${ARCH}-opm ;\
+	curl -sSLo $(OPM) https://github.com/operator-framework/operator-registry/releases/download/v1.19.1/${OS}-${ARCH}-opm ;\
```
abays, line 206: I think the `$$` has special significance in the context of mixed `Makefile` and `bash` syntax.  I think that because the `OS` and `ARCH` variables are set in this recipe as `bash` variables, you might have to reference them with `$$`.  At least, that's what I'm gleaning from reading this: https://stackoverflow.com/a/26564874

PR #34, opened 2022-09-26
Drops the finalizer to delete the DB

dprince:
This drops the DB delete finalizer.

Leaving in the Job and script/template to delete the database so we can follow up with a "DB purge" feature in the future.

### DIFF
--- a/controllers/mariadbdatabase_controller.go
+++ b/controllers/mariadbdatabase_controller.go
@@ -28,7 +28,6 @@ import (
 	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	helper "github.com/openstack-k8s-operators/lib-common/modules/common/helper"
 	job "github.com/openstack-k8s-operators/lib-common/modules/common/job"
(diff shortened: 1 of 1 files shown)


stuggi (commented):
wouldn't it be easier re the follow up purge feature, which I'd really like to have fore dev, to just add the purge parameter now and just add a condition for the dbDeleteJob?

stuggi:
> wouldn't it be easier re the follow up purge feature, which I'd really like to have fore dev, to just add the purge parameter now and just add a condition for the dbDeleteJob?

we could also just have a  different behavior on this between mariadb-operator for dev and the galera one for prod?

dprince:
> wouldn't it be easier re the follow up purge feature, which I'd really like to have fore dev, to just add the purge parameter now and just add a condition for the dbDeleteJob?

We can do the purge feature now. However unlike how the finalizer, the purge feature actually requires changes in all operators in order for it to work correctly. I can add it here however for the time being. And in any case if we do switch things out for a galara-operator it would go away then anyway.

stuggi:
> > wouldn't it be easier re the follow up purge feature, which I'd really like to have fore dev, to just add the purge parameter now and just add a condition for the dbDeleteJob?
> 
> We can do the purge feature now. However unlike how the finalizer, the purge feature actually requires changes in all operators in order for it to work correctly. I can add it here however for the time being. And in any case if we do switch things out for a galara-operator it would go away then anyway.

ack, lets merge it and then do a follow up either in the operator or in install_yamls

stuggi (approved):
/lgtm

PR #35, opened 2022-09-27
Copy generated CRD files dir to /api/bases

stuggi:
Copies the rendered CRD files dir to /api/bases in make manifests when successfully generated.

Also some minor updates for SDK1.23.0


abays (approved):
/lgtm

PR #37, opened 2022-10-05
MariaDB connection redirect for Adoption purposes

jistr:
The MariaDB CR gets a new spec section allowing for temporary redirect to an external DB. Both redirection based on (FQ)DN and based on IP is supported.


api/bases/mariadb.openstack.org_mariadbs.yaml
```diff
@@ -35,6 +35,14 @@ spec:
           spec:
             description: MariaDBSpec defines the desired state of MariaDB
             properties:
+              adoptionRedirect:
```
jistr, line 38: Currently this only has `host` subkey but we may decide to add `port` later too perhaps.

pkg/endpoints.go
```diff
@@ -0,0 +1,36 @@
...
+	"net"
+
+	databasev1beta1 "github.com/openstack-k8s-operators/mariadb-operator/api/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+)
+
+// Endpoints func
+func Endpoints(db *databasev1beta1.MariaDB, scheme *runtime.Scheme) *corev1.Endpoints {
```
stuggi, line 13: seems the scheme is not used? do we need to pass that arg? I guess you only want this in the upper level to set the controller reference?

pkg/service.go
```diff
@@ -5,11 +5,24 @@ import (
...
+	if adoptionHost != "" {
+		if adoptionHostIsIP {
+			return externalServiceFromIP(db, scheme)
+		}
+		return externalServiceFromName(db, scheme)
+	}
+	return internalService(db, scheme)
+}
+
+func internalService(db *databasev1beta1.MariaDB, scheme *runtime.Scheme) *corev1.Service {
```
stuggi, line 25: same, scheme seems not to be used here

pkg/service.go
```diff
@@ -25,3 +38,33 @@ func Service(db *databasev1beta1.MariaDB, scheme *runtime.Scheme) *corev1.Servic
 	}
 	return svc
 }
+
+func externalServiceFromIP(db *databasev1beta1.MariaDB, scheme *runtime.Scheme) *corev1.Service {
```
stuggi, line 42: same, scheme seems not to be used here

pkg/service.go
```diff
@@ -25,3 +38,33 @@ func Service(db *databasev1beta1.MariaDB, scheme *runtime.Scheme) *corev1.Servic
...
+		},
+		Spec: corev1.ServiceSpec{
+			ClusterIP: "None",
+			Type:      corev1.ServiceTypeClusterIP,
+		},
+	}
+	return svc
+}
+
+func externalServiceFromName(db *databasev1beta1.MariaDB, scheme *runtime.Scheme) *corev1.Service {
```
stuggi, line 57: same, scheme seems not to be used here

pkg/endpoints.go
jistr: Scheme seems to be unused in these functions in general, but i wanted to keep a consistent signature with the rest of them, e.g. see the function that creates Pod here:

https://github.com/openstack-k8s-operators/mariadb-operator/blob/597b259525acc6cc36e3653abd9472bf9b91e617/pkg/pod.go#L11

So i think we should either keep the scheme parameter it in all of these "resource object generation" functions or remove them from all? (In another commit, since it's better to not mix feature code with refactorings?)

(In Rust, a linter would force me to name the parameter `_scheme` to mean "this parameter is present in the signature for consistency reasons or interface requirements, but it is not presently used in the function body", but i'm not sure if there is some equivalent of that pattern in Golang...)

stuggi: yeah, that operator needs some cleanup. In general we should not pass not required/used parameters. I am good do the cleanup in a follow up or in the galera replacement operator

pkg/service.go
```diff
@@ -25,3 +38,33 @@ func Service(db *databasev1beta1.MariaDB, scheme *runtime.Scheme) *corev1.Servic
 	}
 	return svc
 }
+
+func externalServiceFromIP(db *databasev1beta1.MariaDB, scheme *runtime.Scheme) *corev1.Service {
+	svc := &corev1.Service{
```
gibizer, line 43: How does this service associated with the external IP?

pkg/endpoints.go
jistr: I added the cleanup into this PR but as a separate commit.

api/v1beta1/mariadb_types.go
```diff
@@ -30,6 +30,14 @@ type MariaDBSpec struct {
 	StorageRequest string `json:"storageRequest,omitempty"`
 
 	ContainerImage string `json:"containerImage,omitempty"`
+
+	AdoptionRedirect AdoptionRedirectSpec `json:"adoptionRedirect,omitempty"`
```
dprince, line 34: Should we tie this only to adoption or call this something more generic like 'serviceEndpoints' or something. Adoption would still use it just the same. Not asking you to change it yet, just bringing it up.

jistr: I considered both options and went for adoption-specific naming because:

* I didn't want to create the impression that customers should expect such features to be supported outside adoption context.

* Having keys named like "adoptionRedirect" or "adoptionDataImport" makes it more clear that:

  * When customer is *not* doing adoption, they shouldn't be setting these.

  * When the adoption procedure is finished and we want the config to converge to how greenfield would look like, these fields should not be present anymore in the CRs. It will make any leftovers easier to spot.

I'm not opposed to generic naming, but i slightly prefer the adoption-specific naming due to the reasons above.

gibizer: For me the naming of this depends on the need to support externally hosted DB outside of the adoption context. Do we have such a need?

pkg/service.go
gibizer: Thanks. Do I understand correctly that the connection between the Sevice and the Endpoints objects is that they both have the same name?

api/v1beta1/mariadb_types.go
dprince: lets leave it as it is then

pkg/service.go
jistr: Yes i think matching name is the link mechanism: https://kubernetes.io/docs/concepts/services-networking/service/#services-without-selectors

gibizer: > When you create an [Endpoints](https://kubernetes.io/docs/reference/kubernetes-api/service-resources/endpoints-v1/) object for a Service, you set the name of the new object to be the same as that of the Service.

OK that explains it. Thanks.

stuggi (approved):
/lgtm

jistr:
Thanks for the reviews :)

PR #41, opened 2022-10-15
Rename the Job used to create the DB

gibizer:
Rename it from x-database-sync to x-db-create as it is not really calling db sync just creating and empty database named x. Also we are not simply dropping underscores from the name but replacing it with a hyphen to make the names more readable.


stuggi (approved):
/lgtm

PR #44, opened 2022-11-28
use stable/zed images

SeanMooney:
This change makes the contianer image optional and defaults
the image to use stable/zed.


stuggi (approved):
/lgtm

PR #45, opened 2022-12-01
[DNM] test

Sandeepyadav93:


Sandeepyadav93:
This was just a test to see unit test job doesn't trigger automatically and is still a required job. We also tried commenting via a non openshift member(to repliacte gh bot commenting) and it works.

PR #48, opened 2023-01-10
Add simple kuttl tests for mariadb

cescgina:
Add a very simple test suite for mariadb that just tests the deployment. This test was extracted from the keystone kuttl test suite and should be expanded in a future change.

### DIFF
--- a/tests/kuttl/tests/mariadb_deploy/01-assert.yaml
+++ b/tests/kuttl/tests/mariadb_deploy/01-assert.yaml
@@ -1,0 +1,62 @@
+#
+# Check for:
+#
+# - 1 MariaDB CR
+# - 1 Pod for MariaDB CR
+#
+
+apiVersion: mariadb.openstack.org/v1beta1
+kind: MariaDB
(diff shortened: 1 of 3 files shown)


tests/kuttl/tests/mariadb_deploy/01-assert.yaml
```diff
@@ -0,0 +1,62 @@
+#
+# Check for:
+#
+# - 1 MariaDB CR
+# - 1 Pod for MariaDB CR
```
stuggi, line 5: could add checking for the service, but could also add that when we have switched to galera as the tests probably will change

stuggi (approved):
/lgtm , as mentioned in the assert file, could add checking for the service, but might leave it for later when switched to galera

stuggi:
like in keystoneapi, should the assert file for the sample always be in `${XYZ_KUTTL_DIR}/../common/assert_keystone_deployment.yaml` so that <service>_deploy_validate in install_yamls always can reference this?

PR #49, opened 2023-01-11
Move deployment assert to a common folder

cescgina:
Move the deployment assert file so that it has the same name for all
operators. The file is then linked inside the kuttl test suite.


stuggi (approved):
/lgtm

PR #50, opened 2023-01-17
Allow attaching service pods to additional networks

stuggi:
For network isolation outgoing traffic, service pods need to be attached additional networks. To achive this, the the `k8s.v1.cni.cncf.io/networks` annotation needs to be set on the pods. This adds a networkAttachmentDefinitions parameter to the CRD where it can be specified to which additional networks the service pods should be attached to.

Depends-On: openstack-k8s-operators/lib-common#133


stuggi:
note, in the current state the mariadb-operator will not restart the mariadb pod when there is a spec change, but I think thats ok for now that its going to be replaced. the pod needs to be deleted to be re-created adding the annotation

stuggi:
closing, for now it is not required to expose the DB to isolated networks

PR #52, opened 2023-01-19
Fix missing DBMaxTimeout template parameter

kajinamit:
The parameter was incompletely removed when we intorduced secrets[1]. This also enables -e so that we can catch any command failures.

[1] d882ed426e3164bd7c0cdd68a78c1dbdd35c161d


dprince:
@kajinamit good catch.

dprince (approved):

PR #66, opened 2023-02-08
Use time.Duration for timeout

kajinamit:
instead of int. This allows us to use more fine grained timeout value such as 0.1, and encodes the unit in the type itself.

Depends-On: openstack-k8s-operators/lib-common#155


kajinamit:
The failure doesn't look relevant to this change ...

```

run golangci-lint
  Running [/home/runner/golangci-lint-1.51.1-linux-amd64/golangci-lint run --out-format=github-actions --timeout 5m] in [] ...
  Error: undefined: RegisterFailHandler (typecheck)
  Error: undefined: RunSpecsWithDefaultAndCustomReporters (typecheck)
  Error: undefined: Reporter (typecheck)
```

gibizer (approved):
looks good

gibizer:
the missing go mod tidy run causing the failures here as well. fixed it now. I expect it to pass CI now.

gibizer (approved):

PR #67, opened 2023-02-09
Fix pre commit

gibizer:
It seems the pre-commit config was merged without running pre-commit on the repo ever. This patch fixes the pre-commit findings.
* setting executable flag for scripts
* adding operator-lint and gowork target and fixing lint findings.


abays (approved):
/lgtm

PR #69, opened 2023-02-13
Finalizers for MariaDBDatabase

abays:
Does two things:

1. Adds a self-finalizer to the `MariaDBDatabase` CR.
2. When a `MariaDBDatabase` CR is created, we now add a finalizer to the associated `Galera` or `MariaDB` to prevent accidental deletion of the latter.  We also remove the finalizer from the `Galera` or `MariaDB` when the `MariaDBDatabase` is deleted.

### DIFF
--- a/controllers/mariadbdatabase_controller.go
+++ b/controllers/mariadbdatabase_controller.go
@@ -28,6 +28,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	helper "github.com/openstack-k8s-operators/lib-common/modules/common/helper"
 	job "github.com/openstack-k8s-operators/lib-common/modules/common/job"
@@ -71,7 +72,7 @@ func (r *MariaDBDatabaseReconciler) GetScheme() *runtime.Scheme {
 // +kubebuilder:rbac:groups=batch,resources=jobs,verbs=get;list;watch;create;update;delete;patch
 
 // Reconcile reconcile mariadbdatabase API requests
-func (r *MariaDBDatabaseReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+func (r *MariaDBDatabaseReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, _err error) {
 	_ = r.Log.WithValues("mariadbdatabase", req.NamespacedName)
 
 	// Fetch the MariaDBDatabase instance
@@ -81,9 +82,35 @@ func (r *MariaDBDatabaseReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
+	helper, err := helper.NewHelper(
+		instance,
+		r.Client,
+		r.Kclient,
+		r.Scheme,
+		r.Log,
+	)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// Always patch the instance status when exiting this function so we can persist any changes.
+	defer func() {
+		err := helper.PatchInstance(ctx, instance)
+		if err != nil {
+			_err = err
+			return
+		}
+	}()
+
+	// If we're not deleting this and the service object doesn't have our finalizer, add it.
+	if instance.DeletionTimestamp.IsZero() && controllerutil.AddFinalizer(instance, helper.GetFinalizer()) {
+		return ctrl.Result{}, nil
+	}
+
 	// Fetch the Galera or MariaDB instance from which we'll pull the credentials
 	// Note: this will go away when we transition to galera as the db
 	var isGalera bool
+	var db client.Object
 	var dbgalera *databasev1beta1.Galera
 	var dbmariadb *databasev1beta1.MariaDB
 	var dbName string
@@ -102,6 +129,8 @@ func (r *MariaDBDatabaseReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		return ctrl.Result{}, err
 	}
 
+	db = dbgalera
+
 	isGalera = err == nil
 	if !isGalera {
 		// Try to fetch MariaDB when Galera is not used
@@ -116,9 +145,15 @@ func (r *MariaDBDatabaseReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		if err != nil && !k8s_errors.IsNotFound(err) {
 			return ctrl.Result{}, err
 		}
+		db = dbmariadb
 	}
 
 	if k8s_errors.IsNotFound(err) {
+		// Handle special case for removing service finalizer in the context of this MariaDBDatabase
+		// being deleted but Galera/MariaDB is not found
+		if !instance.DeletionTimestamp.IsZero() {
+			controllerutil.RemoveFinalizer(instance, helper.GetFinalizer())
+		}
 		return ctrl.Result{RequeueAfter: time.Duration(10) * time.Second}, nil
 	}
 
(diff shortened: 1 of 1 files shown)


controllers/mariadbdatabase_controller.go
```diff
@@ -128,17 +163,33 @@ func (r *MariaDBDatabaseReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+			return ctrl.Result{}, err
+		}
+	}
+
+	// If this MariaDBDatabase is being deleted, there's no reason to continue beyond making
+	// sure the service finalizer has been removed
+	if !instance.DeletionTimestamp.IsZero() {
+		controllerutil.RemoveFinalizer(instance, helper.GetFinalizer())
+		return ctrl.Result{}, nil
 	}
```
gibizer, line 187: I think the finalizer logic is correct but I'd like to simplify this finalizer implementation a bit.

```golang
db, err := GetDB(...) // get the MariaDB or Galera CR as proposed

// if we are being deleted then we have to remove the finalizer from MariaDB/Galera and then remove it from ourselves
if !instance.DeletionTimestamp.IsZero() {
    if err != nil {  // so we have MariaDB or Galera to remove finalizer from
        if controllerutil.RemoveFinalizer(db, fmt.Sprintf("%s-%s", helper.GetFinalizer(), instance.Name)) {
            err := r.Update(ctx, db)
            if err != nil {
                return ctrl.Result{}, err
            }
        }    
    }
    
    // all our external cleanup logic is done so we can remove our own finalizer to signal that we can be deleted.
    controllerutil.RemoveFinalizer(instance, helper.GetFinalizer())
    // we can unconditionally return here as this is basically the end of the delete sequence for MariaDBDatabase
    // so nothing else needs to be done in the reconcile.
    return ctrl.Result{}, nil
}

// we now that this is not a delete case

        // as it is not a delete case we need to wait for MariaDB or Galera to exists before we can continue.
	return ctrl.Result{RequeueAfter: time.Duration(10) * time.Second}, nil
}

// here we know that MariaDB or Galera exists so add a finalizer to ourselves and to the db CR. Before this point there is no reason to have a finalizer on ourselves as nothing to cleanup.

if instance.DeletionTimestamp.IsZero() {    // this condition can be removed if you wish as it is always true at this point otherwise we would returned earlier.
    if controllerutil.AddFinalizer(db, fmt.Sprintf("%s-%s", helper.GetFinalizer(), instance.Name)) {
        err := r.Update(ctx, db)
        if err != nil {
            return ctrl.Result{}, err
        }
    }

    if controllerutil.AddFinalizer(instance, helper.GetFinalizer()) {
        // we need to persist this right away
        return ctrl.Result{}, nil
    }
}

// we can do our normal flow
```
On positive thing of this code that it has clearer separation between the delete and the non delete codepath. I can even suggest to pull a function out for the delete path to further separate them.

controllers/mariadbdatabase_controller.go
```diff
@@ -116,9 +145,15 @@ func (r *MariaDBDatabaseReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
+		db = dbmariadb
 	}
 
 	if k8s_errors.IsNotFound(err) {
+		// Handle special case for removing service finalizer in the context of this MariaDBDatabase
+		// being deleted but Galera/MariaDB is not found
+		if !instance.DeletionTimestamp.IsZero() {
+			controllerutil.RemoveFinalizer(instance, helper.GetFinalizer())
+		}
 		return ctrl.Result{RequeueAfter: time.Duration(10) * time.Second}, nil
```
gibizer, line 157: I think we can end up here when no MariaDB or Galera exists and MariaDBDatabase is being deleted. I'm not sure what will happen if we requeue here. Will this be an infinite requeue as we never responded that it is OK to be deleted? Or as soons as the finalizer is removed from ourselves the framework deletes the CR and stops reconciling?

gibizer (changes requested):

controllers/mariadbdatabase_controller.go
abays: Since `!instance.DeletionTimestamp.IsZero()`, k8s is thus trying to delete the instance but can't yet due to the finalizer.  So once we remove the finalizer, k8s will continue to delete the instance.

controllers/mariadbdatabase_controller.go
```diff
@@ -116,9 +145,15 @@ func (r *MariaDBDatabaseReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		if err != nil && !k8s_errors.IsNotFound(err) {
 			return ctrl.Result{}, err
 		}
+		db = dbmariadb
 	}
 
 	if k8s_errors.IsNotFound(err) {
+		// Handle special case for removing service finalizer in the context of this MariaDBDatabase
+		// being deleted but Galera/MariaDB is not found
```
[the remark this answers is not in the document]
abays: It shouldn't be a common execution path by any means.  But someone could, for instance, delete the `MariaDB` instance and then manually remove the finalizer (via `oc edit` or whatever), and then you'd be in this situation.

controllers/mariadbdatabase_controller.go
abays: Sure, I can refactor it like this

controllers/mariadbdatabase_controller.go
gibizer: OK, so it CR will be deleted even if there a reconcile is requeued for it.

abays: I will test it again to make sure my understanding is correct, but I'm pretty sure it's fine

controllers/mariadbdatabase_controller.go
abays: Actually, it can't be quite as simple as this.  The "normal flow" needs two additional things:

1. Some way of indicating that either `Galera` or `MariaDB` is being processed
2. The actual `Galera` or `MariaDB` object

So perhaps the return of the proposed `GetDB()` func can return more than just `client.Object` and `error`.  I'll look into it.

controllers/mariadbdatabase_controller.go
```diff
@@ -128,17 +163,33 @@ func (r *MariaDBDatabaseReconciler) Reconcile(ctx context.Context, req ctrl.Requ
...
-		r.Client,
-		r.Kclient,
-		r.Scheme,
-		r.Log,
-	)
-	if err != nil {
-		return ctrl.Result{}, err
+	// Either add or remove the MariaDBDatabase finalizer for this instance to/from the associated Galera/MariaDB
+	// instance based on whether this MariaDBDatabase is being deleted or not
+	if instance.DeletionTimestamp.IsZero() && controllerutil.AddFinalizer(db, fmt.Sprintf("%s-%s", helper.GetFinalizer(), instance.Name)) {
```
stuggi, line 168: should the finalizer always be more specific then just the `gvk.Kind`? https://github.com/openstack-k8s-operators/lib-common/blob/master/modules/common/helper/helper.go#L75 could be changed to have this format per default.

abays: I suppose we could do that.  In this case it's not a "self-finalizer" though, but a finalizer placed on another object.  But we could always use a more nuanced finalizer no matter what the target.

controllers/mariadbdatabase_controller.go
gibizer: ahh you are right. I ignored the details of GetDB. But adding GetDB wasn't my main point anyway. So I'm OK to keep that inline.

controllers/mariadbdatabase_controller.go
gibizer: For self-finalizer kind probably works. But it does not hurt to be more specific there either. So I'm totally OK to change the default finalizer name to <kind>-<name> format. So the result of h.GetFinalizer() can always be used by the instance as a finalizer on itself or on any other type.

controllers/mariadbdatabase_controller.go
```diff
@@ -71,89 +72,122 @@ func (r *MariaDBDatabaseReconciler) GetScheme() *runtime.Scheme {
...
-			return ctrl.Result{}, err
-		}
-	}
+	}()
 
-	if k8s_errors.IsNotFound(err) {
-		return ctrl.Result{RequeueAfter: time.Duration(10) * time.Second}, nil
+	// If we're not deleting this and the service object doesn't have our finalizer, add it.
+	if instance.DeletionTimestamp.IsZero() && controllerutil.AddFinalizer(instance, helper.GetFinalizer()) {
+		return ctrl.Result{}, nil
```
gibizer, line 109: you can drop this as L153 already handles it

gibizer (commented):
one cleanup and then I think ti is good to go

controllers/mariadbdatabase_controller.go
abays: Ack

gibizer (approved):
Looks good. Thanks @abays

PR #78, opened 2023-02-20
Fix mariadb config file permissions and restart pod on config change

olliewalsh:
Mariadb is using default conf as the conf file is not readable by the mysql user. Also pass the config hash as an env to the pod to ensure it is recreated when the config changes.

### DIFF
--- a/templates/mariadb/bin/config.json
+++ b/templates/mariadb/bin/config.json
@@ -4,7 +4,7 @@
         {
             "source": "/var/lib/config-data/galera.cnf",
             "dest": "/etc/my.cnf.d/galera.cnf",
-            "owner": "root",
+            "owner": "mysql",
             "perm": "0600"
         }
     ]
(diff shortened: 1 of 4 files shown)


templates/mariadb/bin/config.json
```diff
@@ -4,7 +4,7 @@
         {
             "source": "/var/lib/config-data/galera.cnf",
             "dest": "/etc/my.cnf.d/galera.cnf",
-            "owner": "root",
+            "owner": "mysql",
```
olliewalsh, line 7: @dciabrin is this also an issue for the galera controller?

dciabrin: Yes the galera pod is started as root as well and mysqld switches to user mysql, so the owner of this file is still root.

dciabrin: Heads up, in TripleO all files under /etc/my.cnf.d are owned by root, and this config is also owned by root with  perms 644.
I think this is what we want here as well, and I did the same for the galera CR.

PR #86, opened 2023-03-29
mariadb: allow mysqld to set max_connections at startup

dciabrin:
Mariadb pods are started as user mysql, which has no specific right to configure file descriptor limits. This prevents the server from honoring the requested max_connections of 4096.

Make the pod start mysqld as root, and let the mysqld server drop privileges to user mysql explicitly. This way, the server can configure limits prior to accepting connections.

### DIFF
--- a/pkg/pod.go
+++ b/pkg/pod.go
@@ -9,6 +9,7 @@ import (
 // Pod -
 func Pod(db *databasev1beta1.MariaDB, configHash string) *corev1.Pod {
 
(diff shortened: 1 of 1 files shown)


gibizer (commented):
Looks good to me. We agreed that this is a temporary solution until we switch to galera by default

bogdando (commented):
/lgtm

stuggi:
@dciabrin want to rebase?

PR #87, opened 2023-03-31
galera: fix permissions of config files

dciabrin:
Since [1] the kolla config incorrectly sets permissions of the mysql config directory to be accessible by root only. Fix read access for all users, and align the permissions on the config files as well.

[1] 751b510968ab82fe5dd61b2198fb5cd30df0427e


stuggi (approved):
/lgtm

PR #88, opened 2023-04-03
Add kuttl test case deploying galera

cescgina:
Add a kuttl test case deploying galera and checking that all the
resources are created and deleted, and the deployment can be scaled
down.


tests/kuttl/tests/galera_deploy/01-assert.yaml
```diff
@@ -0,0 +1,205 @@
...
+    cr: galera-openstack
+---
+apiVersion: v1
+kind: Endpoints
+metadata:
+  name: openstack-galera
+  namespace: openstack
+---
+apiVersion: v1
+data:
```
dciabrin, line 105: Could we instead only check whether the config-map is created only?
I'm not sure we want to tweak this test any time we need to update the value of the configmap's keys

cescgina:
@dciabrin thanks for the review. It seems that tide requires either the `/approve` label or approving the PR through the github interface to merge, could you do that?

PR #91, opened 2023-04-05
Fix make tidy not to depend on ./bin path

gibizer:
We should not assume that ./bin exists as it is not part of the git repository but created dynamically. Without this renovate might fail to generate a PR as ./bin does not exists in a fresh clone of the repo.


gibizer:
The make tidy target was fixed independently so only the renovate change remains.

stuggi (approved):
/lgtm

PR #93, opened 2023-04-07
galera: improve recovery of pods

dciabrin:
The mariadb operator determines whether a starting pod will bootstrap a galera cluster or join an existing one. On startup, the galera pod receives that information via a temporary file injected by the operator. Once galera is running in the pod, the operator updates its status.

Improve the status tracking so that the operator can detect when a pod failed to start galera, and try again in a subsequent reconcile loop.

Also make sure that the operator cannot inspect the galera state while the galera server is being started, to prevent any corruption.

### DIFF
--- a/controllers/galera_controller.go
+++ b/controllers/galera_controller.go
@@ -59,6 +59,10 @@ type GaleraReconciler struct {
 	Scheme  *runtime.Scheme
 }
 
+///
+// General Galera helper functions
+//
+
 // findBestCandidate returns the node with the lowest seqno
 func findBestCandidate(status *mariadbv1.GaleraStatus) string {
 	sortednodes := maps.Keys(status.Attributes)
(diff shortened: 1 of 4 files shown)


controllers/galera_controller.go
```diff
@@ -452,69 +605,6 @@ func (r *GaleraReconciler) generateConfigMaps(
...
-func isBootstrapInProgress(instance *mariadbv1.Galera) bool {
-	for _, attr := range instance.Status.Attributes {
-		if attr.Gcomm == "gcomm://" {
-			return true
-		}
-	}
-	return false
-}
-
-// isBootstrapInProgress checks whether a node is currently starting a galera cluster
```
dprince, line 503: comment here should say removeOldAttributesOnScaleDown

dprince (approved):

PR #101, opened 2023-04-19
galera: do not require root user to init/run pods

dciabrin:
The galera pods are configured to run as root so that Kolla can copy config files and bootstrap a database. Later, mysqld drops privileges by switching to user mysql via setuid().

Do not explicitely require the root user and use sudo rights instead. This works because Kolla is already allowed to sudo, and the mysqld server can set limits (e.g. file descriptors, sockets) without requiring root privileges.

### DIFF
--- a/templates/galera/bin/mysql_probe.sh
+++ b/templates/galera/bin/mysql_probe.sh
@@ -1,23 +1,27 @@
 #!/bin/bash
-set -eux
+set -eu
 
 # This secret is mounted by k8s and always up to date
 read -s -u 3 3< /var/lib/secrets/dbpassword MYSQL_PWD || true
 export MYSQL_PWD
 
+PROBE_USER=root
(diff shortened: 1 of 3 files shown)


templates/galera/bin/mysql_probe.sh
```diff
@@ -1,23 +1,27 @@
 #!/bin/bash
-set -eux
+set -eu
```
dprince, line 2: any reason not to include the -x here?

dciabrin: I noticed that when a probe fails, k8s logs an event with an excerpt of what the probe logged on stderr. Since the start probe fails until mysql is started, I wanted the event to log the connection failure without being cloggeed with bash's debug output.

dprince (approved):

PR #102, opened 2023-04-20
Dynamic serviceaccount/role/rolebinding creation

dprince:
This PR dyamically creates the above resources within the
namespace of the requested MariaDB/Galera instance.

The above resources are no longer created via OLM.

This allows the operator to service requests in any
namespace within the OpenShift cluster.

Also includes changes to support exec.StreamWithContext

This resolves golangci errors due to k8s go.mod updates and was
required to make the above pass.


dprince:
missing some logic to set the new rbac conditions to true

dprince:
The kuttl failure above is due to the fact that infra-operator relies (incorrectly, a bug) on the mariadb-operator-mariadb service account. This PR fixes the issue:
https://github.com/openstack-k8s-operators/infra-operator/pull/32

tests/kuttl/common/assert_sample_deployment.yaml
```diff
@@ -32,6 +32,18 @@ status:
     reason: Ready
     status: "True"
     type: MariaDBInitialized
+  - message: RoleBinding not created
+    reason: Init
+    status: Unknown
+    type: RoleBindingReady
```
dprince, line 38: This is actually wrong. I put in a fix to lib-common here: https://github.com/openstack-k8s-operators/lib-common/pull/244

abays (approved):
/lgtm

PR #103, opened 2023-04-27
Use automatically generated namespace for kuttl tests

cescgina:
After recent changes in dynamic service account, since mariadb can be
deployed in any namespace, this change modifies the way the kuttl tests
are run, so they run in a namespace automatically created (and deleted) by
kuttl. This allows us to run the kuttl tests in a cluster where we have
already a deployed instance, without interfering. Additionally, we get
the benefit of being able to run multiple tests in parallel, since each
runs in its own namespace, which also reduces the runtime.

### DIFF
--- a/kuttl-test.yaml
+++ b/kuttl-test.yaml
@@ -18,8 +18,7 @@ apiVersion: kuttl.dev/v1beta1
 kind: TestSuite
 reportFormat: JSON
 reportName: kuttl-test-mariadb
-namespace: openstack
 timeout: 180
-parallel: 1
+parallel: 2
 suppress:
   - events                     # Remove spammy event logs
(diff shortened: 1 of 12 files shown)


abays:
I've cloned the branch used here and tried to use it via:

```
MARIADB_KUTTL_CONF=/home/ocp/podified/cescgina-mariadb-operator/kuttl-test.yaml MARIADB_KUTTL_DIR=/home/ocp/podified/cescgina-mariadb-operator/tests/kuttl/tests NAMESPACE=baystest make mariadb_kuttl
```

But I don't see any resources being created in the `baystest` namespace.  Would you mind sharing your testing steps for this?  Thanks!

abays (approved):
/lgtm

PR #104, opened 2023-05-04
normalize DEFAULT_IMAGE and kustomization.yaml

yazug:
add DEFAULT_IMAGE to Makefile
Also update config/manager/kustomization.yaml for what is generated currently


stuggi (approved):
/lgtm

PR #113, opened 2023-05-30
Remove unused status.safeToBootstrap

kajinamit:
Galera CR exposes status.safeToBootstrap but this field is never updated. This change removes the unused field.


dciabrin:
We want to keep this one for future state tracking during bootstrap when we will parse the safe_to_bootstrap flag out of galera config directly.

kajinamit:
I'm ok with keeping this field if we have an actual plan to implement it.

In case we don't have specific timeline for the feature implementation then I prefer removing it for now to avoid exposing non-fuctioning field. We can add it back when we implement the logic.

PR #119, opened 2023-06-28
Support TLS connections for galera

dciabrin:


controllers/galera_controller.go
```diff
@@ -393,6 +409,36 @@ func (r *GaleraReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
...
+	}
+	if err == nil && tls != nil && tls.Ca.CaSecretName != "" {
+		_, caHash, err = secret.GetSecret(ctx, helper, tls.Ca.CaSecretName, instance.Namespace)
+	}
+	if k8s_errors.IsNotFound(err) {
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			condition.InputReadyCondition,
+			condition.RequestedReason,
+			condition.SeverityInfo,
+			condition.InputReadyWaitingMessage))
```
gibizer, line 429: It would help troubleshooting if this message contains information about which secret is missing.

controllers/galera_controller.go
```diff
@@ -473,16 +519,43 @@ func (r *GaleraReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
...
 	// This is used in an envvar in the statefulset to restart it on config change
-	envHash := &corev1.EnvVar{}
-	configMapVars[configMapNameForConfig(instance)](envHash)
-	instance.Status.ConfigHash = envHash.Value
+
+	keys := make([]string, 0, len(configMapVars))
+	for k := range configMapVars {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
```
gibizer, line 542: this can be simplified to:
```golang
keys := maps.Keys(configMapVars)
sort.Strings(keys)
```
https://pkg.go.dev/golang.org/x/exp/maps#Keys

pkg/mariadb/database.go
```diff
@@ -14,12 +14,23 @@ type dbCreateOptions struct {
 	DatabaseName          string
 	DatabaseHostname      string
 	DatabaseAdminUsername string
+	DatabaseUserTLS       string
 }
 
 // DbDatabaseJob -
 func DbDatabaseJob(database *databasev1beta1.MariaDBDatabase, databaseHostName string, databaseSecret string, containerImage string, serviceAccountName string) (*batchv1.Job, error) {
+	return DbDatabaseJobExt(database, databaseHostName, databaseSecret, containerImage, serviceAccountName, false)
+}
```
gibizer, line 23: We probably don't need this backward compatibility adapter for the internal part of mariadb-operator. (If this is a function in api/v1beta1 then compatibility might be needed)

controllers/galera_controller.go
dciabrin: Good idea, done

pkg/mariadb/database.go
dciabrin: You're right, it's internal only, so I extended the original function as you suggested.

controllers/galera_controller.go
dciabrin: Done

controllers/mariadbdatabase_controller.go
```diff
@@ -147,6 +148,11 @@ func (r *MariaDBDatabaseReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		dbSecret = dbGalera.Spec.Secret
 		dbContainerImage = dbGalera.Spec.ContainerImage
 		serviceAccount = dbGalera.RbacResourceName()
+		// TODO(dciabrin) DB user can be configured to only allow connections to the DB
+		// over TLS. Do not enforce this until client support for TLS connection is
+		// implemented.
+		// useTLS = (dbGalera.Spec.TLS.SecretName != "")
+		useTLS = false
```
gibizer, line 155: Do we need to switch to the commented out code to be able to test this?

gibizer (commented):
overall looks good to me

controllers/mariadbdatabase_controller.go
dciabrin: I updated it slightly to enforce TLS for users based on TLS property "TLS.Service.DisableNonTLSListeners". I also added a kuttl test for that feature

dciabrin:
I just updated kuttl tests to add a test for each TLS feature included in this PR.

controllers/galera_controller.go
```diff
@@ -663,5 +737,22 @@ func (r *GaleraReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+				}
+
+				return []reconcile.Request{
+					{NamespacedName: types.NamespacedName{
+						Name:      reconcileCR,
+						Namespace: o.GetNamespace(),
+					}},
+				}
+			},
+		)).
```
stuggi, line 756: is the label being set somewhere ? you consider using it to watch the tls cert secrets? for this case we intend that the service operators add themselves as an additional owner (not controller) of the secret. then is should only be required to add `Owns(&corev1.Secret{}).`

controllers/galera_controller.go
```diff
@@ -44,15 +45,22 @@ import (
...
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	mariadbv1 "github.com/openstack-k8s-operators/mariadb-operator/api/v1beta1"
 	mariadb "github.com/openstack-k8s-operators/mariadb-operator/pkg/mariadb"
 )
 
+// Label used in a k8s secret to reference its corresponding galera CR
+const mariaDBReconcileLabel = "mariadb-ref"
```
stuggi, line 62: is it added somewhere to an object?

controllers/galera_controller.go
```diff
@@ -473,16 +523,40 @@ func (r *GaleraReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
...
+
+	keys := maps.Keys(configMapVars)
+	sort.Strings(keys)
+
+	hash := ""
+	for _, k := range keys {
+		envVar := &corev1.EnvVar{}
+		configMapVars[k](envVar)
+		hash = hash + envVar.Value
+	}
```
stuggi, line 550: could do similar to other services where we create a hash of hashes? https://github.com/openstack-k8s-operators/keystone-operator/blob/main/controllers/keystoneapi_controller.go#L755 / https://github.com/openstack-k8s-operators/keystone-operator/blob/main/controllers/keystoneapi_controller.go#L1118

controllers/galera_controller.go
dciabrin: No the idea was to rely on the templating ability of cert-manager, as e.g. https://github.com/openstack-k8s-operators/mariadb-operator/pull/119/files#diff-c73cb33ebf19bd8e12d56f29679841e5abcd55b1fb58b07ec416749e8ca2d30bR47-R49
The drawback is that any Secret won't work if it's not labeled.
I'm good with adding the label automatically if this is a pattern that we're going to follow everywhere in the project.

controllers/galera_controller.go
dciabrin: See my remark below

controllers/galera_controller.go
dciabrin: Ack I'll update the review accordingly.

controllers/galera_controller.go
stuggi: lets wait with this. I should have an example hopefully tomorrow or on thu to show what I mean

controllers/galera_controller.go
dciabrin: Done in the latest revision. We now also keep a hash of all the individual inputs for easier tracking of changed hashes

controllers/galera_controller.go
dciabrin: I just implemented the 2nd pattern in this PR, that uses cache indexing. This approach allows us to reconcile and restart galera when a secret is manually deleted to forcibly regenerate a certificate in cert-monger.

dciabrin:
Fixed unit tests in the last push

pkg/mariadb/volumes.go
```diff
@@ -113,3 +114,181 @@ func getInitVolumeMounts() []corev1.VolumeMount {
...
+		}, {
+			MountPath: "/var/lib/operator-scripts",
+			ReadOnly:  true,
+			Name:      "operator-scripts",
+		}, {
+			MountPath: "/var/lib/kolla/config_files",
+			ReadOnly:  true,
+			Name:      "kolla-config",
+		},
+	}
```
olliewalsh, line 291: Does this need the TLS volume mounts too?

api/bases/mariadb.openstack.org_galeras.yaml
```diff
@@ -60,6 +60,10 @@ spec:
                   defaults, or overwrite rendered information using raw MariaDB config
                   format. The content gets added to /etc/my.cnf.d/galera_custom.cnf
                 type: string
+              disableNonTLSListeners:
```
Deydra71, line 63: We don't support this in tls package anymore, you can safely delete this ^^

config/crd/bases/mariadb.openstack.org_galeras.yaml
```diff
@@ -60,6 +60,10 @@ spec:
                   defaults, or overwrite rendered information using raw MariaDB config
                   format. The content gets added to /etc/my.cnf.d/galera_custom.cnf
                 type: string
+              disableNonTLSListeners:
```
Deydra71, line 63: ^^

api/bases/mariadb.openstack.org_galeras.yaml
dciabrin: Ack, I piggybacked that flag to enforce TLS requirements for database user creation. When this is in use, all database users created via MariaDBDatabase CR will always enforce TLS. There is no per-user TLS support yet, hence me resurrecting it.
I can rename it to be more meaningful if you prefer?

Deydra71: Ahhh, I see. Then it all makes sense to me. Thanks for explaining!

api/v1beta1/galera_types.go
```diff
@@ -66,6 +67,13 @@ type GaleraSpec struct {
 	// +kubebuilder:validation:Optional
 	// Adoption configuration
 	AdoptionRedirect AdoptionRedirectSpec `json:"adoptionRedirect"`
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec
+	// TLS settings for MySQL service and internal Galera replication
+	TLS tls.SimpleService `json:"tls,omitempty"`
+	// +kubebuilder:validation:Optional
+	// When TLS is used, only allow connections to the DB over TLS
```
olliewalsh, line 75: nit: maybe s/used/configured/

controllers/galera_controller.go
```diff
@@ -97,7 +120,8 @@ func buildGcommURI(instance *mariadbv1.Galera) string {
 	res := []string{}
 
 	for i := 0; i < replicas; i++ {
-		res = append(res, basename+"-"+strconv.Itoa(i)+"."+basename)
+		// Generate Gcomm with FQDN for TLS validation
```
olliewalsh, line 123: nit: it not the FQDN, that would have .cluster.local on the end (logic looks fine, just the comment)

pkg/mariadb/volumes.go
```diff
@@ -25,91 +69,141 @@ func getVolumes(name string) []corev1.Volume {
...
-			Name: "lib-data",
-			VolumeSource: corev1.VolumeSource{
-				PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
-					ClaimName: "mariadb-" + name,
-				},
-			},
-		},
 	}
 
+	if g.Spec.TLS.Enabled() {
```
stuggi, line 118: not super important, but I think we could simplify this a bit like in https://github.com/openstack-k8s-operators/nova-operator/blob/main/pkg/novncproxy/deployment.go#L128-L135 ?

pkg/mariadb/volumes.go
```diff
@@ -25,91 +69,141 @@ func getVolumes(name string) []corev1.Volume {
...
 			Name:      "kolla-config",
 		},
-		{
-			MountPath: "/var/lib/mysql",
-			ReadOnly:  false,
-			Name:      "lib-data",
-		},
 	}
 
+	if g.Spec.TLS.Enabled() {
```
stuggi, line 163: same as above

controllers/galera_controller.go
```diff
@@ -470,9 +501,51 @@ func (r *GaleraReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
...
+	if err != nil {
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			condition.InputReadyCondition,
+			condition.ErrorReason,
+			condition.SeverityWarning,
+			condition.InputReadyErrorMessage,
+			err.Error()))
+		return ctrl.Result{}, fmt.Errorf("error calculating input hash: %w", err)
+	}
+	instance.Status.Conditions.MarkTrue(condition.InputReadyCondition, condition.InputReadyMessage)
```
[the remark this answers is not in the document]
dciabrin: Ack I'll update that

pkg/mariadb/volumes.go
dciabrin: That one was on purpose because ToService() can return an error which would make me update the current function's signature and its caller, for something that look purely static to me.
I can update that if you feel that would look more in line with other operators

pkg/mariadb/volumes.go
dciabrin: No, the init is only running the database to initialize innodb, and we explicitly use a default config that doesn't depend on TLS for that

api/v1beta1/galera_types.go
dciabrin: done

controllers/galera_controller.go
dciabrin: done

dciabrin:
PR rebased to add TLS support on the new MariaDBAccount CR as well

pkg/mariadb/volumes.go
stuggi: ack, its fine

stuggi (approved):
/lgtm

PR #122, opened 2023-07-05
Temporarily use 1 Galera replica as a workaround

abays:
We're doing this while we investigate a MariaDB 10.5 issue with replication


abays:
This will not help us after all

PR #123, opened 2023-07-06
Watch RBAC resources to trigger reconcile

kajinamit:
This ensures the controller watches service account, role, and role bindings to reconcile these resources in case any change is made.


kajinamit:

looks like timeout during image build ?

stuggi (approved):
/lgtm

PR #125, opened 2023-07-12
Remove duplicate jobs from github workflow

kajinamit:
We now run pre-commit job in prow which covers most of go-lang tasks.


kajinamit:
It's wiered that these removed jobs are still required. I'm wondering if rebase helps here...

PR #127, opened 2023-07-19
Configure renovate to pin k8s and controller-runtime for dev preview

gibizer:
This pins k8s and controller runtime to be in sync with lib-common and configure renovate to keep the pin


stuggi (approved):
/lgtm

PR #130, opened 2023-07-20
Use pre-created osp-secret in kuttl tests

kajinamit:
Currently we create osp-secret in kuttl test but this conflicts with the secret created by install_yamls. This removes the creation of the secret in kuttl to avoid updating the existing secret.
    
This also updates the sample file to use that common secret, so that we can use the file in kuttl tests without modification.

Depends-on: https://github.com/openstack-k8s-operators/install_yamls/pull/425


tests/kuttl/tests/galera_deploy/03-cleanup_galera.yaml
```diff
@@ -1,7 +1,12 @@
+apiVersion: kuttl.dev/v1beta
+kind: TestStep
+delete:
```
cescgina, line 3: This `delete` should be in the same TestStep as the script below, kuttl does not allow to have two `TestStep` blocks in the same file. In the docs there is an example of using the delete and commands fields in the same TestStep: https://kuttl.dev/docs/testing/reference.html#teststep

kajinamit: Fixed in the latest version.

kajinamit:
I tested this with https://github.com/openstack-k8s-operators/install_yamls/pull/425 .

abays (approved):
/lgtm

PR #138, opened 2023-07-31
Use lib-common for pvc

rabi:
Looks like mariadb pvcs are not cleaned up when the mariadb instances are deleted. Let's use the lib-common methods to CreateOrPatch pvc. Whether to retain the pv when pvc are deleted would be decided by the StorageClass.

### DIFF
--- a/controllers/mariadb_controller.go
+++ b/controllers/mariadb_controller.go
@@ -25,7 +25,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/api/resource"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


kajinamit:
I discussed this with @dciabrin a while ago and the conclusion was that this is an expected behavior to avoid wiping out data in case the instance is deleted unintentionally. but I'd like Damien to double-check my understanding.

rabi:
> I discussed this with @dciabrin a while ago and the conclusion was that this is an expected behavior to avoid wiping out data in case the instance is deleted unintentionally.

Not data would be wiped out if a `storageClass` with `reclaimPolicy: Retain` is used. If we want to be different and want to ensure that we keep the data irrespective of the `storageClass` used, we've to make it consistent across all operators for example ovndb is deleted[1] when service is deleted.

[1] https://github.com/openstack-k8s-operators/ovn-operator/blob/main/pkg/ovndbcluster/statefulset.go#L163

kajinamit:
> 

Technically ovn db can be rebuilt from neutron db(though we'd lost all data inserted by octavia db), but I agree this is something we should set via reclaimPolicy in StorageClass instead of leaving pvc.

rabi:
> It sounds like there is maybe a concern here in using the shared lib-common routines in that we really want to guarantee database's aren't accidentally deleted

Is this is a concern only for mariadb? Isn't `reclaimPolicy` the k8s way of not accidentally deleting disk? We need to consistent across services if we want to deviate from that IMO.

PR #140, opened 2023-08-01
Drop redundant configMapNameForConfig/Scripts

kajinamit:
These functions contain only a single line logic and are used in only a single place.


kajinamit:
This is still used in one more place... Probably we should create a method in pkg and use it.

PR #142, opened 2023-08-03
switch to structured logging

pinikomarov:
This automatically adds additional fields like reconcile_id etc.. from the controller context.

before :

`2023-05-18T01:53:14+03:00 INFO  controllers.KeystoneAPI Reconciled Service init successfully`

after:

`2023-10-19T10:38:14+03:00       INFO    Controllers.galera      StatefulSet openstack-cell1-galera - updated    {"controller": "galera", "controllerGroup": "mariadb.openstack.org", "controllerKind": "Galera", "Galera": {"name":"openstack-cell1","namespace":"openstack"}, "namespace": "openstack", "name": "openstack-cell1", "reconcileID": "ceb646eb-6400-460a-a49a-a75b8c821abc"}
`

*by using per reconcile function respective logger objects, the intention is to lay the ground for parallel reconciliation in future and avoid race conditions as ctx objects are reconcile run specific.

### DIFF
--- a/controllers/galera_controller.go
+++ b/controllers/galera_controller.go
@@ -58,8 +58,12 @@ type GaleraReconciler struct {
 	client.Client
 	Kclient kubernetes.Interface
 	config  *rest.Config
-	Log     logr.Logger
 	Scheme  *runtime.Scheme
+}
+
+// GetLog returns a logger object with a prefix of "controller.name" and additional controller context fields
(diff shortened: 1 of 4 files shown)


pinikomarov:
@abays @dciabrin does somebody needs to set "/approved" label here so this can continue to merge ?

abays (approved):
/lgtm

PR #144, opened 2023-08-11
galera: support mariabackup SST method

dciabrin:
Make the SST method configurable in the galera custom resource, and allow both rsync and mariabackup methods.
Mariabackup requires a database user's credentials to operate, so reuse the root db user for the time being.

The ability to inject passwords in template galera config files has been implemented. We benefit from that to generate the usual .my.cnf config file for mysql CLI running in the galera container.


dprince (approved):

stuggi:
@dciabrin this is fairly old, you want to rebase and get this in, or is it no longer required?

dciabrin:
@stuggi this is still neeting but it needs rework.

PR #155, opened 2023-09-27
Move database test helpers from lib-common to /api

gibizer:
This is necessary to remove a dependency cycle


gibizer:
/hold 
I want to test his

PR #163, opened 2023-10-20
[testhelpers] Add helpers for Galera (create, delete, get)

stuggi:


gibizer (approved):
Looks good. I'm just wondering where will we use these. Currently service operators only depend on the k8s Service object (see CreateDBService) to exists during envtest testing.

stuggi:
> Looks good. I'm just wondering where will we use these. Currently service operators only depend on the k8s Service object (see CreateDBService) to exists during envtest testing.

I plan to use it in the openstack-operator where we create the db instances

gibizer:
> > Looks good. I'm just wondering where will we use these. Currently service operators only depend on the k8s Service object (see CreateDBService) to exists during envtest testing.
> 
> I plan to use it in the openstack-operator where we create the db instances

Ohh, good point. I will be happy to review envtest coverage for openstack-operator!

PR #166, opened 2023-10-30
add controller to create and delete individual usernames in mariadb / galera

zzzeek:
this is a "create /drop account" controller that is separate from the main "create/drop database" controller, for the purpose of producing rotating username/passwords.    The background for the change is based on discussions surrounding https://issues.redhat.com/browse/OSPRH-92 where internal control plane services such as Galera , Rabbit, Redis etc. would provide interfaces to add /remove arbitrary usernames, where a "password rotation" would involve adding a new username/password and having services switch there, retiring the old account once all finalizers have been removed.


api/v1beta1/mariadbaccount_types.go
```diff
@@ -0,0 +1,75 @@
...
+	Items           []MariaDBAccount `json:"items"`
+}
+
+func init() {
+	SchemeBuilder.Register(&MariaDBAccount{}, &MariaDBAccountList{})
+}
+
+const (
+	// UserNameKey - key in secret which holds the account username
+	UserNameKey = "UserName"
```
zzzeek, line 74: this is not used now, remove

zzzeek (commented):
we should add a sample file that includes a MariaDBAccount and a Secret

zzzeek:
not sure if I should keep this in draft, the folks I wanted to see it first were @dciabrin , @SeanMooney and @gibizer

api/v1beta1/mariadbaccount_types.go
```diff
@@ -0,0 +1,70 @@
...
+	// UserName for new account
+	UserName string `json:"userName,omitempty"`
+
+	// Name of secret which contains DatabasePassword
+	Secret string `json:"secret,omitempty"`
+}
+
+// MariaDBAccountStatus defines the observed state of MariaDBAccount
+type MariaDBAccountStatus struct {
+	Completed bool `json:"completed,omitempty"`
```
gibizer, line 42: We tend to use Ready Condition to signal that the operation is completed. But if you only want to express a single flag then the boolean is enough. I'm wondering if you need to expose some error message when the reconciliation fails. If so then a Condition would be better.

controllers/mariadbaccount_controller.go
```diff
@@ -0,0 +1,322 @@
...
+// +kubebuilder:rbac:groups=mariadb.openstack.org,resources=MariaDBAccounts,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=mariadb.openstack.org,resources=MariaDBAccounts/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=mariadb.openstack.org,resources=MariaDBAccounts/finalizers,verbs=update
+// +kubebuilder:rbac:groups=mariadb.openstack.org,resources=mariadbs/status,verbs=get;list
+// +kubebuilder:rbac:groups=mariadb.openstack.org,resources=galeras/status,verbs=get;list
+// +kubebuilder:rbac:groups=batch,resources=jobs,verbs=get;list;watch;create;update;delete;patch
+
+// Reconcile reconcile MariaDBAccount API requests
+func (r *MariaDBAccountReconciler) Reconcile(ctx context.Context, req ctrl.Request) (result ctrl.Result, _err error) {
+	_ = r.Log.WithValues("MariaDBAccount", req.NamespacedName)
```
gibizer, line 62: I would use this log all over the place as the framework already set it up to contain relevant informations like the name and uuid of the instance being reconciled. See https://github.com/openstack-k8s-operators/placement-operator/pull/94

controllers/mariadbaccount_controller.go
```diff
@@ -0,0 +1,322 @@
...
+	}()
+
+	isDelete := !instance.DeletionTimestamp.IsZero()
+
+	// locate the MariaDBDatabase object that this account is associated with
+	mariadbDatabase, err := r.getMariaDBDatabaseObject(ctx, instance)
+
+	// if not found because the label is missing, don't requeue, will be
+	// reconciled again when label is fixed
+	if err != nil && k8s_errors.IsNotFound(err) && instance.ObjectMeta.Labels["mariaDBDatabaseName"] == "" {
```
gibizer, line 99: this condition seems strange as it mixing the NotFound error an the missing label case.

controllers/mariadbaccount_controller.go
```diff
@@ -0,0 +1,322 @@
...
+		r.Log.Info(fmt.Sprintf("MariaDBAccount '%s' does not have a 'mariaDBDatabaseName' label; returning", instance.Name))
+		return ctrl.Result{}, err
```
gibizer, line 101: I think this can cause that if a MariaDBAccount is created without the label then it cannot be deleted as the reconciler will return here with an error.

controllers/mariadbaccount_controller.go
```diff
@@ -0,0 +1,322 @@
...
+					"MariaDBAccount '%s' MariaDBDatabase '%s' not yet complete; no account delete needed",
+					instance.Name, instance.ObjectMeta.Labels["mariaDBDatabaseName"]))
+			}
+			controllerutil.RemoveFinalizer(instance, helper.GetFinalizer())
+
+			return ctrl.Result{}, nil
+		}
+	} else if err != nil {
+		// unhandled error; exit
+		return ctrl.Result{}, nil
```
gibizer, line 140: For troubleshooting purposes please at least log the error but also probably better to return the error

controllers/mariadbaccount_controller.go
```diff
@@ -0,0 +1,322 @@
...
+			return ctrlResult, nil
+		}
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		if accountDeleteJob.HasChanged() {
+			if instance.Status.Hash == nil {
+				instance.Status.Hash = make(map[string]string)
+			}
+			// TODO: do we set a hash for deletes?
```
gibizer, line 264: If you want to avoid double deletion cases then yes you need to persist the hash

zzzeek:

hold-the-node

main.go
SeanMooney, line 1: before thsi is ready for eview i assume you are going to do an interactive rebase and remove the intemeriday commits correct?

config/crd/bases/mariadb.openstack.org_mariadbaccounts.yaml
SeanMooney, line 1: this file shoudl be generated form the API class definition using kubebuilder annotations

like is done here https://github.com/openstack-k8s-operators/mariadb-operator/blob/main/api/v1beta1/mariadb_types.go#L32-L49

so you should not be creating this manually or in a separate commit to the one that adds the golang class

SeanMooney (changes requested):
im conflicted on how to procedd with reviewing this.

we are not meant to use sqash merges in any of the repos which means you should not be breaking out the commits the way you are.

each commit should be indepently correct and when you need to make fixes you should interactivly rebase the fixes into the approate commits.

so I'm not sure there is value in doing an review of the implementation until that is done

i can take a look at the design side i guess but can you try cleaning this set of commits up in the 3-5 that it probably should be.

i would start with having the first commit introduce the API type/CRD
then the controller and samples
followed by kuttle tests and then ideally envtest
i realise that the mariadb-operator currently does not have envtest coverage which is why that should be last and likely in a sepeata pr since it will require some infra to be set up in the repo first

main.go
zzzeek: I can do that sure.  I usually work with Gerrit where none of that is an issue.   was not sure what the convention is for Github PRs.   want me to just commit --amend  + push -f for every change?

config/crd/bases/mariadb.openstack.org_mariadbaccounts.yaml
zzzeek: yah i slowly figured out which files get generated vs. not vs. "generated but you have to edit it" as i went through so i can re-do more of the files based on that

zzzeek: to be clear, I consider this to still be "draft" but I wanted to get the CI tests to pass and also get guidance on practices here, like "use the generator tools here" and "squash the commits".    That the workflow seems to be implying that you can't "review" it yet indicates I'm either not getting the workflow yet, or there is a shortcoming that there's no space for "not ready for final review but please correct my track here"

zzzeek: this file was actually generated.  maybe in an earlier commit it was a cut and paste job, so anyway, the squash will make it clearer what we are looking at

controllers/mariadbaccount_controller.go
zzzeek: right, I can just check for the missing label earlier all by itself and be done with it.

config/crd/bases/mariadb.openstack.org_mariadbaccounts.yaml
zzzeek: I would note that there is no kuttl test for the existing MariaDBDatabase resource, to which MariaDBAccount depends.    

so do we need a separate PR first for MariaDBDatabase kuttl tests added?    Or I can have that be part of the stream here.

zzzeek:

not sure if this works in draft mode, I want to get CI going first before I add all the Condition stuff

main.go
SeanMooney: you can have multiple commits in a series but if you need to modify content that was introduced in an earlier commit i would personally do that via an interactive rebase and commit --amend then push the full series at the end ya.

breaking thing out into logical chunks still makes sense in the github flow but you should treat it like gerrit where unless its a a comment typo fix or something that does not impact the funcaitonlty i.e. a style issue  you should fix the commits in place with ammed rather then adding follow up fixes.

controllers/galera_controller.go
```diff
@@ -665,3 +667,45 @@ func (r *GaleraReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&rbacv1.RoleBinding{}).
 		Complete(r)
 }
+
+// GetDatabaseObject - returns either a Galera or MariaDB object (and an associated client.Object interface).
+// used by both MariaDBDatabaseReconciler and MariaDBAccountReconciler
+// this will later return only Galera objects, so as a lookup it's part of the galera controller
+
+func GetDatabaseObject(clientObj client.Client, ctx context.Context, name string, namespace string) (client.Object, *databasev1beta1.Galera, *databasev1beta1.MariaDB, error) {
```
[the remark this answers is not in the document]
zzzeek: this is an existing function that I moved out of the MariaDBDatabase controller, so that it can be used by both.  Should I also change MariaDBDatabase controller to use the new return value in this PR ?

controllers/mariadbaccount_controller.go
zzzeek: Trying out this log, this does the logging differently than what we have in the other parts of this controller:

new way:

```
2023-11-30T12:40:08-05:00       INFO    MariaDBDatabase 'neutron' is ready      {"controller": "mariadbaccount", "controllerGroup": "mariadb.openstack.org", "controllerKind": "MariaDBAccount", "MariaDBAccount": {"name":"neutron1","namespace":"openstack"}, "namespace": "openstack", "name": "neutron1", "reconcileID": "e409f2c8-3fe2-45a3-bc31-d10fccd6adf4", "MariaDBAccount": "openstack/neutron1"}

```

old way:

```
2023-11-30T12:40:08-05:00       INFO    controllers.Galera      Galera mikegalera database headless service mikegalera-galera - operation: updated
2023-11-30T12:40:08-05:00       INFO    controllers.Galera      Galera mikegalera database service mikegalera - operation: updated

```

so this also would seem like something that is more cleanly applied to all controllers at once in a separate PR, do you agree?

controllers/galera_controller.go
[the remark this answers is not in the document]
zzzeek: also comments in the code suggest the "mariadb" portion of the controller is going away anyway, so that would also fix this problem when that happens.  @dciabrin can you confirm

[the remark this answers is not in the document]
zzzeek: I just updated both mariadbdatabase and mariadbaccount controllers to receive back the Client, it uses a runtime type check (new Go skill unlocked) to cast to one or the other

api/v1beta1/mariadbaccount_types.go
zzzeek: ive replaced with Condition and a few sub-conditions.  we can continue to tune how that's done since this looks a little like it intends to look a certain way inside the status, but general idea is there

controllers/mariadbaccount_controller.go
zzzeek: ive moved the missing label check to be separate.

controllers/mariadbaccount_controller.go
```diff
@@ -0,0 +1,400 @@
...
+	// things we are setting up are named in the spec"
+	mariadbDatabaseName := instance.ObjectMeta.Labels["mariaDBDatabaseName"]
+	if mariadbDatabaseName == "" {
+		r.Log.Info(fmt.Sprintf("MariaDBAccount '%s' does not have a 'mariaDBDatabaseName' label; returning", instance.Name))
+		return ctrl.Result{}, err
+	}
+	// locate the MariaDBDatabase object that this account is associated with
+	mariadbDatabase, err := r.getMariaDBDatabaseObject(ctx, instance, mariadbDatabaseName)
+
+	r.Log.Info("mariadb datrabase object")
```
zzzeek, line 141: typo, but also this looks like an unnecessary log line

controllers/mariadbaccount_controller.go
zzzeek: OK I assume we can't try to guess for the case where someone modifies the labels in a maridbaccount object, then deletes it.

so from this it would follow that all "delete" operations need to be made to proceed here despite any other validation errors, like mariadbaccount was stated, but not found.

api/v1beta1/mariadbaccount_types.go
```diff
@@ -17,25 +17,34 @@ limitations under the License.
...
+	Secret string `json:"secret,omitempty"`
 }
 
 // MariaDBAccountStatus defines the observed state of MariaDBAccount
 type MariaDBAccountStatus struct {
-	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
+	// Deployment Conditions
+	Conditions condition.Conditions `json:"conditions,omitempty" optional:"true"`
+
```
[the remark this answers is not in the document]
zzzeek: as it stands, this new secret would always be identical to the one given, we aren't generating passwords here.  we had said the Secret would be immutable to start with, or did you mean we'd *output* a secret with a possibly autogen-password and *that's* what's immutable?   I didnt have the impression we were doing autogen of passwords here

api/bases/mariadb.openstack.org_mariadbaccounts.yaml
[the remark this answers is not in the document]
SeanMooney: this is just the partern that @gibizer followed in the nova operator when we were creating the operator from scratch.

the only docs related to this we have is https://github.com/openstack-k8s-operators/docs/blob/main/new_operator.md#create-templates-folder-holding-scriptsconfig-templates
which does not go into group changes into commits.
https://github.com/openstack-k8s-operators/nova-operator/pull/61/commits

you do not need to follow that process exactly.

api/v1beta1/mariadbaccount_types.go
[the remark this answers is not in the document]
zzzeek: ok graet.  yes I proposed autogeneration originally but we all seemed to decide we didnt want that here. in any case autogen password within the controller could be added in a separate PR

api/bases/mariadb.openstack.org_mariadbaccounts.yaml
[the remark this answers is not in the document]
zzzeek: that PR seems to have the controllers only in machine-generated state, I will add some more commits here to separate out the machine generated vs human generated

api/v1beta1/mariadbaccount_types.go
```diff
@@ -17,25 +17,34 @@ limitations under the License.
...
 type MariaDBAccountSpec struct {
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
+	// UserName for new account
+	UserName string `json:"userName,omitempty"`
 
-	// Foo is an example field of MariaDBAccount. Edit mariadbaccount_types.go to remove/update
-	Foo string `json:"foo,omitempty"`
+	// Name of secret which contains DatabasePassword
+	Secret string `json:"secret,omitempty"`
```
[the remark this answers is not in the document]
zzzeek: the required validation marks are added

controllers/mariadbaccount_controller.go
zzzeek: we're going to try to merge #142 first, I can rebase on that and update logging here

api/v1beta1/mariadbaccount_types.go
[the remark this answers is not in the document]
SeanMooney: that is not the expecte behaivor. the expect behavior is that if the secret does not exist we would set a input validation condition to  false and in the reason state that the secret is missing then the operator should return without executing the CreateDBAccountJob.

i would consider it a bug to attpet to execute any ansible EE job or any API request against the galara/mariadb instance betore doing the input validation to assert the required secrete is present and that it has the filed we expect.

we have some helper function to do the secret existence and content check in lib-common.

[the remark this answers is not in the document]
zzzeek: okey doke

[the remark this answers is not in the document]
zzzeek: what if Secret exists but missing the key? same thing?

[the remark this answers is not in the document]
zzzeek: nm I foudn the util will use that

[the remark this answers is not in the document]
zzzeek: done

controllers/mariadbaccount_controller.go
gibizer: In case of create without a label we return here with success (ctrl.Result{}, err=nil) but that is not an issue as when the label is added to the Accont CR that will cause a reconcile to be triggered as it is a change on the CR.

controllers/mariadbaccount_controller.go
```diff
@@ -0,0 +1,451 @@
...
+			// our own instance and return
+			if err != nil {
+				r.Log.Info(fmt.Sprintf(
+					"MariaDBAccount '%s' Didn't find MariaDBDatabase '%s'; no account delete needed",
+					instance.Name, instance.ObjectMeta.Labels["mariaDBDatabaseName"]))
+
+			} else {
+				r.Log.Info(fmt.Sprintf(
+					"MariaDBAccount '%s' MariaDBDatabase '%s' not yet complete; no account delete needed",
+					instance.Name, instance.ObjectMeta.Labels["mariaDBDatabaseName"]))
```
gibizer, line 199: In this case do we have a Finalizer on the MariaDBDatabase instance? If so we need to remove that as well.
// later
I see below that  we only add Finalizer to the DB instance after it is Completed state. So the question is can that instance go to Completed state, get a finalizer, and then go back to an incomplete state? If so then we need to remove finalizer from the DB here as it might have it. Fortunately RemoveFinalizer will not fail if you try to remove a finalizer that is not exist.

controllers/mariadbaccount_controller.go
```diff
@@ -0,0 +1,451 @@
...
+
+		return ctrl.Result{}, err
+	}
+
+	var dbInstance, dbAdminSecret, dbContainerImage, serviceAccountName string
+
+	// ensure Galera/MariaDB instance itself is ready
+	// TODO: not sure to what extent galera / mariadb would not be ready
+	// at this point, considering MariaDBDatabase is dependent on this state
+	// as well and that's been assured as "ready"
```
gibizer, line 273: I think it is OK to rely on the readyness of the MariaDBDatabase and assume that it can only be ready if the underlying Galera instance is ready. If it is not true then I would fix the condition handling of MariaDBDatabase instead of building extra logic here.

controllers/mariadbaccount_controller.go
```diff
@@ -0,0 +1,451 @@
...
+			[]string{"DatabasePassword"},
+			r.Client,
+			time.Duration(30)*time.Second,
+		)
+		if err != nil {
+			instance.Status.Conditions.Set(condition.FalseCondition(
+				databasev1beta1.MariaDBAccountReadyCondition,
+				secret.ReasonSecretMissing,
+				condition.SeverityInfo,
+				databasev1beta1.MariaDBAccountReadyInitMessage))
```
gibizer, line 324: It would be super useful during troubleshooting to include the error into the condition message to tell the user which field is missing from the Secret

controllers/mariadbaccount_controller.go
```diff
@@ -0,0 +1,451 @@
...
+		// remove finalizer from our own instance
+		controllerutil.RemoveFinalizer(instance, helper.GetFinalizer())
+
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	return ctrl.Result{}, nil
+}
```
gibizer, line 419: `!isDelete` conditions are sprinkled all over the Reconcile function. I'm wondering if it will be easier to follow the two main codepath if they would be separated from the start into two independent function instead of interleaving it into a single function. This can be a refactor after this PR when we have enough test coverage that we can use to prove that such refactor is functionally equivalent.

controllers/mariadbaccount_controller.go
```diff
@@ -0,0 +1,451 @@
...
+				databasev1beta1.MariaDBAccountReadyCondition,
+				secret.ReasonSecretMissing,
+				condition.SeverityInfo,
+				databasev1beta1.MariaDBAccountReadyInitMessage))
+
+			return result, err
+		}
+
+		r.Log.Info(fmt.Sprintf("Running account create '%s' MariaDBDatabase '%s'", instance.Name, mariadbDatabaseName))
+
```
SeanMooney, line 330: so here before you call mariadb.CreateDbAccountJob
you should validate if dbAdminSecret exists and has the expected fields 

you can do that with the VerifySecret function form lib common

https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/secret/secret.go#L410-L447

that was proted form a sligly more complex version we had in the nova operator.
we have not replaced our usage of our custom version with this yet but we can look at the dataplane operator for an example fo how to use it

https://github.com/openstack-k8s-operators/dataplane-operator/blob/7e706e92418811cd2625cb3a77c3c3fad2156cd2/controllers/openstackdataplanenodeset_controller.go#L232-L265

something like this

```
	secretKeys := []string{"password"}
	_, result, err = secret.VerifySecret(
		ctx,
		types.NamespacedName{
			Namespace: instance.Namespace,
			Name:      dbAdminSecret,
		},
		secretKeys,
		helper.GetClient(),
		time.Second*5,
	)

	if err != nil {
		if (result != ctrl.Result{}) {
			instance.Status.Conditions.MarkFalse(
				condition.InputReadyCondition,
				condition.RequestedReason,
				condition.SeverityInfo,
				dataplanev1.InputReadyWaitingMessage,
				"secret/"+dbAdminSecret)
		} else {
			instance.Status.Conditions.MarkFalse(
				condition.InputReadyCondition,
				condition.RequestedReason,
				condition.SeverityWarning,
				err.Error())
		}
		return result, err
	}
```

zzzeek: hi, mentioned in my [comment](https://github.com/openstack-k8s-operators/mariadb-operator/blob/c4a46ae4c77e53bfe7711684f3509989d4af304e/controllers/mariadbaccount_controller.go#L311) this is implemented nearly identically to what you have there (did you miss it?), though I can add the extra conditional for the result thing you have there


https://github.com/openstack-k8s-operators/mariadb-operator/blob/e765d04ccca5f82c31d8635b8c51f02ef1f44077/controllers/mariadbaccount_controller.go#L311-L331

controllers/mariadbaccount_controller.go
zzzeek: yes I think because i have the conditionals nested in this way for two different "not found " / "not ready" statuses, and then I test the conditional a second time inside each block, this whole block should be broken out into two top levels so that this mistake is easier to catch

controllers/mariadbaccount_controller.go
SeanMooney: nope github just sucks for code review and i missed that line propaly because of the if !delete

so this is fine as is.

controllers/mariadbaccount_controller.go
zzzeek: so for Galera, it definitely can go into a "not bootstrapped" state even though there's a full database there, I'm not sure to what extent the galera controller here is linking "Bootstrapped" to the ongoing state of Galera but as this is just a single "if not bootstrapped...", it seems both safe and correct to have it there.

for the MariaDB non-galera case, my understanding is that this whole case is going away at some point and we will just have the "galera" version, not the "mariadb" version.

so,...i will remove my "todo" comment and just leave these two conditionals in, IMO it's better to check the condition, log it , and handle it, rather than crash if those states were not set up.

controllers/mariadbaccount_controller.go
zzzeek: yah I looked at VerifySecret and it always returns `ctrl.Result{}` for the result field, so perhaps at some point in the past it had different result scenarios.   I dont think I would change what I'm doing for a different result type so I will just make the error condition here more detailed.

zzzeek: whoops ignore that comment

controllers/mariadbaccount_controller.go
SeanMooney: we are free to drop the usage and support of mariadb today.
all operators are expected to support galara today.

we have one or two examples that should be updated when we do remove it

such as https://github.com/openstack-k8s-operators/openstack-operator/blob/main/config/samples/core_v1beta1_openstackcontrolplane.yaml#L18-L36

but that is a change that can be made when we have time to do so.

i.e. if you had completed your q4 work and had time left then you could do this basically now. in reality i think probably should get planned to do in the new year.

controllers/mariadbaccount_controller.go
zzzeek: i've migrated to the new logging

controllers/mariadbaccount_controller.go
zzzeek: > In this case do we have a Finalizer on the MariaDBDatabase instance? If so we need to remove that as well.

this should be fixed now

> I see below that we only add Finalizer to the DB instance after it is Completed state. So the question is can that instance go to Completed state, get a finalizer, and then go back to an incomplete state? 

MariaDBDatabase is like you ran "CREATE DATABASE nova" on the server, so it works similarly to this controller; the resource is considered to be immutable once created.

so I'm not sure yet what it would mean for MariaDBDatabase to move from Complete to "incomplete", or why that implies mariadbaccount needs to remove its finalizer for that case.

controllers/mariadbaccount_controller.go
zzzeek: that will be great, once I get this merged it will be quick

controllers/mariadbaccount_controller.go
zzzeek: this is formatted as `databasev1beta1.MariaDBAccountSecretNotReadyMessage, err` in the call

templates/account.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+export DatabasePassword=${DatabasePassword:?"Please specify a DatabasePassword variable."}
+
+mysql -h {{.DatabaseHostname}} -u {{.DatabaseAdminUsername}} -P 3306 -e "GRANT ALL PRIVILEGES ON {{.DatabaseName}}.* TO '{{.UserName}}'@'localhost' IDENTIFIED BY '$DatabasePassword';GRANT ALL PRIVILEGES ON {{.DatabaseName}}.* TO '{{.UserName}}'@'%' IDENTIFIED BY '$DatabasePassword';"
```
olliewalsh, line 4: Need to control TLS here too?

zzzeek: hmmm do you mean how we are connecting to the server here or like a "REQUIRE TLS" for the user account itself?  the format of the script here I am copying from the existing [database.sh](https://github.com/openstack-k8s-operators/mariadb-operator/blob/main/templates/database.sh) script which isn't doing that right now.   if we were doing "REQUIRE TLS" I'd assume that would be something configurable in the CRD and could also be done in a followup PR if we are not sure yet.   @dciabrin do you have a suggestion?

dciabrin: I have the TLS bits in my PR that add TLS for galera and gcomm. Right now this PR doesn't have support for TLS so I'd keep them separated and if the TLS PR merges first, we'll adapt Mike's PR.

zzzeek: sounds good.

dprince:
I will let @dciabrin merge it but I think it looks good what you have here

dprince:
@zzzeek one thing I noticed locally with this. Can you run 'make bundle' and commit the modifications to config/manifests/bases/mariadb-operator.clusterserviceversion.yaml as part of this PR?

zzzeek:
> @zzzeek one thing I noticed locally with this. Can you run 'make bundle' and commit the modifications to config/manifests/bases/mariadb-operator.clusterserviceversion.yaml as part of this PR?

OK will do that now, hopefully it makes it all the way through

zzzeek:
ok thats up now, just added an entry to a yaml file

zzzeek:
OK so I also have to add this API to the PROJECT file, is that generated from something ?

zzzeek:
operator-sdk did it already

controllers/mariadbaccount_controller.go
```diff
@@ -0,0 +1,451 @@
...
+	client.Client
+	Kclient kubernetes.Interface
+	Log     logr.Logger
+	Scheme  *runtime.Scheme
+}
+
+// SetupWithManager -
+func (r *MariaDBAccountReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&databasev1beta1.MariaDBAccount{}).
```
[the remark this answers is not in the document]
gibizer: I think we cannot require it to be immutable from the consumer side. Where we create the Secret we can set the field ``immutable: true```[1] and then k8s makes sure that the content of the Secret cannot change. 

[1] https://kubernetes.io/docs/concepts/configuration/secret/#secret-immutable

controllers/mariadbaccount_controller.go
gibizer: OK. The added RemoveFinalizer call solves this.

controllers/mariadbaccount_controller.go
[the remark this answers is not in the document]
zzzeek: we can make this operator actually check that the secret is immutable and reject the operation if it's not.  Or in theory this operator could modify the secret itself to be immutable (since that operation is allowed) but I'm assuming we dont want a data mutation going in that direction.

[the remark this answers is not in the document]
zzzeek: yah I will add a condition for it so it shows up in the cr status field, that seems to be the best way to communicate with the "user"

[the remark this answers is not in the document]
gibizer: Yes. We tend to have an InputReady condition[1] in other operators to signal if the user input is incomplete (e.g. a field is missing from a Secret provided to us)

[1]https://github.com/openstack-k8s-operators/lib-common/blob/f7a552f208e7981290c98b7f08c4ae078eadd4a3/modules/common/condition/conditions.go#L26-L27

controllers/mariadbaccount_controller.go
```diff
@@ -0,0 +1,553 @@
...
+		err := r.Update(ctx, mariadbDatabase)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	if controllerutil.AddFinalizer(instance, helper.GetFinalizer()) {
+		// we need to persist this right away
+		return ctrl.Result{}, nil
+	}
```
gibizer, line 217: nit: I would reorder this two steps. Adding the finalizer to our instance signals that there might be things to cleanup during delete. And one of the things to cleanup, removing the finalizer from the mariadbDatabase during delete. This way we can avoid the (unlikely scenario) that we added a finalizer to the mariadbDatabase, got killed (e.g. k8s decided to move the controller pod), the instance gets a delete request from outside and k8s happily deletes it as there are no finalizer on it, then the controller comes up but does not found the instance so does nothing, so the finalizer on the mariadbDatabase got leaked.

controllers/mariadbaccount_controller.go
```diff
@@ -0,0 +1,553 @@
...
+		log.Info(fmt.Sprintf(
+			"MariaDBAccount '%s' MariaDBDatabase '%s' not yet complete; no account delete needed",
+			instance.Name, instance.ObjectMeta.Labels["mariaDBDatabaseName"]))
+
+		controllerutil.RemoveFinalizer(instance, helper.GetFinalizer())
+
+		// remove finalizer from the MariaDBDatabase instance
+		if controllerutil.RemoveFinalizer(mariadbDatabase, fmt.Sprintf("%s-%s", helper.GetFinalizer(), instance.Name)) {
+			err = r.Update(ctx, mariadbDatabase)
+		}
```
gibizer, line 390: I would reorder this. 
1) do the cleanup
2) remove the finalizer from the dependency
3) remove the finalizer from ourselves. 
See my above comment about a potential race.

controllers/mariadbaccount_controller.go
```diff
@@ -0,0 +1,553 @@
...
+		log.Info(fmt.Sprintf("Job %s hash added - %s", jobDef.Name, instance.Status.Hash[databasev1beta1.AccountDeleteHash]))
+	}
+
+	// remove finalizer from the MariaDBDatabase instance
+	if controllerutil.RemoveFinalizer(mariadbDatabase, fmt.Sprintf("%s-%s", helper.GetFinalizer(), instance.Name)) {
+		err = r.Update(ctx, mariadbDatabase)
+	}
+
+	// remove finalizer from our own instance
+	controllerutil.RemoveFinalizer(instance, helper.GetFinalizer())
```
gibizer, line 518: This is in the correct order.

gibizer (commented):
This looks good I'm OK to land this and do my remaining comments in a follow up.

controllers/mariadbaccount_controller.go
zzzeek: it's absolutely a good point, ill add that now

controllers/mariadbaccount_controller.go
zzzeek: yah same thing.  this one is less critical though since it isnt returning a result in between

controllers/mariadbaccount_controller.go
zzzeek: yah.   this "nesting" problem is literally a core SQLAlchemy thing as well re: inserting / deleting dependent rows, creating / dropping dependent tables, so I should know better :)

zzzeek:
> This looks good I'm OK to land this and do my remaining comments in a follow up.

OK it should just be ensuring the secret is immutable change subsequent to this

gibizer:
/lgtm
Thanks for the quick respin!

PR #171, opened 2023-11-16
fix bundle validate: Add Galera to config/samples

dprince:
Resolves:

WARN[0000] Warning: Value mariadb.openstack.org/v1beta1, Kind=Galera: provided API should have an example annotation


stuggi (approved):
/lgtm

PR #172, opened 2023-11-23
Make database character encoding configurable

dciabrin:
When a MariaDBDatabase CR is reconciled, a database is created with a default character encoding and collation of latin1.

Instead of falling back to the server's default encoding for database creation, expose the encoding in the MariaDBDatabase CRD and default to utf8, as it expected by various Openstack projects.

### DIFF
--- a/config/crd/bases/mariadb.openstack.org_mariadbdatabases.yaml
+++ b/config/crd/bases/mariadb.openstack.org_mariadbdatabases.yaml
@@ -35,11 +35,23 @@ spec:
           spec:
             description: MariaDBDatabaseSpec defines the desired state of MariaDBDatabase
             properties:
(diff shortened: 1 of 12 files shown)


beagles (approved):

config/crd/bases/mariadb.openstack.org_mariadbdatabases.yaml
```diff
@@ -35,11 +35,23 @@ spec:
           spec:
             description: MariaDBDatabaseSpec defines the desired state of MariaDBDatabase
             properties:
+              defaultCharacterSet:
+                default: utf8
```
zzzeek, line 39: how come we went with utf8 and not utf8mb4 ? https://mariadb.com/kb/en/unicode/

PR #173, opened 2023-12-06
Remove MariaDB CRD

zzzeek:
openstack-k8s-operators now uses Galera exclusively for MySQL/MariaDB access, so the MariaDB CRD can be fully removed.


config/rbac/role.yaml
```diff
@@ -177,33 +177,13 @@ rules:
...
-- apiGroups:
-  - mariadb.openstack.org
-  resources:
-  - mariadbs/finalizers
-  verbs:
-  - update
 - apiGroups:
   - mariadb.openstack.org
   resources:
   - mariadbs/status
```
zzzeek, line 183: hmm this should not be here

zzzeek:
OK it looks like there are references to this from both install_yamls as well as openstack-operator, so I can't quite trace out how the build is failing but we would assume the openstack-operator build is somehow failing because things it expects to see have been removed from here.

zzzeek:
this should be all set now

api/v1beta1/galera_types.go
```diff
@@ -18,14 +18,24 @@ package v1beta1
...
+
+	// GaleraContainerImage is the fall-back container image for Galera
+	GaleraContainerImage = "quay.io/podified-antelope-centos9/openstack-mariadb:current-podified"
 )
 
+// AdoptionRedirectSpec defines redirection to a different DB instance during Adoption
+type AdoptionRedirectSpec struct {
+	// MariaDB host to redirect to (IP or name)
+	Host string `json:"host,omitempty"`
+}
```
gibizer, line 37: @jistr Does adoption depends on this today or it was an old remnant from the early days of adoption planning? (at least I don't see it mentioned in the data-plane-adoption repo)

gibizer (commented):
I'm wondering if this test helper needs to be change as well:
https://github.com/openstack-k8s-operators/mariadb-operator/blob/4cd7b3f5eedec8fef56396cfb88f3cbd795dbb08/api/test/helpers/crd.go#L103C11-L152 
as it refers to the code that looks for MariaDB CR by name 
https://github.com/openstack-k8s-operators/mariadb-operator/blob/4cd7b3f5eedec8fef56396cfb88f3cbd795dbb08/api/v1beta1/mariadbdatabase_funcs.go#L79-L87

zzzeek:
> I'm wondering if this test helper needs to be change as well: https://github.com/openstack-k8s-operators/mariadb-operator/blob/4cd7b3f5eedec8fef56396cfb88f3cbd795dbb08/api/test/helpers/crd.go#L103C11-L152 as it refers to the code that looks for MariaDB CR by name
> 
> https://github.com/openstack-k8s-operators/mariadb-operator/blob/4cd7b3f5eedec8fef56396cfb88f3cbd795dbb08/api/v1beta1/mariadbdatabase_funcs.go#L79-L87

cc @dciabrin does this comment also apply to the Galera CR ?

api/v1beta1/galera_types.go
zzzeek: so this cannot be trivially removed right now without changes to the galera_controller itself, as it has a lot of logic that refers to the AdoptionRedirect element of the datastructure, it also implies removal of endpoints.go and all that.  Since galera_controller.go would be impacted by this can I propose removing this be a separate PR since this is not really part of the "remove MariaDB" theme of this change.

gibizer: @zzzeek sure lets do the cleanup in a separate PR.

dprince:
> > I'm wondering if this test helper needs to be change as well: https://github.com/openstack-k8s-operators/mariadb-operator/blob/4cd7b3f5eedec8fef56396cfb88f3cbd795dbb08/api/test/helpers/crd.go#L103C11-L152 as it refers to the code that looks for MariaDB CR by name
> > https://github.com/openstack-k8s-operators/mariadb-operator/blob/4cd7b3f5eedec8fef56396cfb88f3cbd795dbb08/api/v1beta1/mariadbdatabase_funcs.go#L79-L87
> 
> cc @dciabrin does this comment also apply to the Galera CR ?

I think you are correct. The test helper will need to be updated. But the breakage wouldn't occur until renovate bumps up the mariadb-operator/api dependency right? So we could land this and gradually update the operators to use an equivalent Galera test helper gradually

zzzeek:
> > > I'm wondering if this test helper needs to be change as well: https://github.com/openstack-k8s-operators/mariadb-operator/blob/4cd7b3f5eedec8fef56396cfb88f3cbd795dbb08/api/test/helpers/crd.go#L103C11-L152 as it refers to the code that looks for MariaDB CR by name
> > > https://github.com/openstack-k8s-operators/mariadb-operator/blob/4cd7b3f5eedec8fef56396cfb88f3cbd795dbb08/api/v1beta1/mariadbdatabase_funcs.go#L79-L87
> > 
> > 
> > cc @dciabrin does this comment also apply to the Galera CR ?
> 
> I think you are correct. The test helper will need to be updated. But the breakage wouldn't occur until renovate bumps up the mariadb-operator/api dependency right? So we could land this and gradually update the operators to use an equivalent Galera test helper gradually

still trying to understand the complete flow of this helper, I'm not yet seeing the "breakage" here since this helper seems to be in a closed loop against the local mariadb API in any case (comments are out of date since https://github.com/openstack-k8s-operators/lib-common/pull/352 occurred) but this seems to raise a bigger issue which is that the `mariadbv1.Database` API is what all the services are using to create databases, and that's not yet integrated with the new MariaDBAccount thing I just merged, seems like it should be.

dciabrin:
> > > > I'm wondering if this test helper needs to be change as well: https://github.com/openstack-k8s-operators/mariadb-operator/blob/4cd7b3f5eedec8fef56396cfb88f3cbd795dbb08/api/test/helpers/crd.go#L103C11-L152 as it refers to the code that looks for MariaDB CR by name
> > > 
> > > cc @dciabrin does this comment also apply to the Galera CR ?
> > 

Labeling shouldn't change, the Galera CR and the MariaDB CR both use the same labels so clients looking for the database service don't need to treat MariaDB and Galera separately. Back when this was introduced, the only component that would distinguish those two CR was the MariaDBDatabase controller.
So I would say this labeling should all be internal, and no change is needed. If we want to revisit later and replace references to mariadb by galera, this can be done in another PR.

zzzeek:
I've deployed openstack locally using this version of the operator, there's no MariaDB crd in the cluster at all and everything is running fine

```
$ oc get crds | grep mariadb
galeras.mariadb.openstack.org                                     2024-01-11T16:21:22Z
mariadbaccounts.mariadb.openstack.org                             2024-01-11T16:21:22Z
mariadbdatabases.mariadb.openstack.org                            2024-01-11T16:21:22Z

[classic@hab-11 install_yamls]$ oc get pods
NAME                           READY   STATUS    RESTARTS      AGE
ceilometer-0                   3/3     Running   0             24m
cinder-api-0                   2/2     Running   0             20m
cinder-scheduler-0             2/2     Running   0             20m
dnsmasq-dns-69cb44859d-gcjgb   1/1     Running   0             16m
glance-default-single-0        3/3     Running   0             24m
keystone-6d94d9cd4-794x5       1/1     Running   0             23m
memcached-0                    1/1     Running   0             30m
neutron-84f6b4848c-gcjwz       2/2     Running   0             17m
nova-api-0                     2/2     Running   0             15m
nova-cell0-conductor-0         1/1     Running   0             17m
nova-cell1-conductor-0         1/1     Running   0             15m
nova-cell1-novncproxy-0        1/1     Running   0             17m
nova-metadata-0                2/2     Running   0             15m
nova-scheduler-0               1/1     Running   0             15m
openstack-cell1-galera-0       1/1     Running   0             30m
openstack-galera-0             1/1     Running   0             30m
openstackclient                1/1     Running   0             21m
ovn-controller-5gzth           3/3     Running   0             30m
ovn-northd-79cf7987f5-54jjw    1/1     Running   0             27m
ovsdbserver-nb-0               1/1     Running   0             30m
ovsdbserver-sb-0               1/1     Running   0             30m
placement-7444b6685d-nhf8k     2/2     Running   0             21m
rabbitmq-cell1-server-0        1/1     Running   0             30m
rabbitmq-server-0              1/1     Running   1 (27m ago)   30m
swift-proxy-549bcf8f4d-t9jmh   3/3     Running   0             24m
swift-storage-0                16/16   Running   0             30m


```

controllers/mariadbaccount_controller.go
```diff
@@ -416,7 +398,7 @@ func (r *MariaDBAccountReconciler) reconcileDelete(
 	// referenced by the MariaDBDatabase which will lead us to the hostname
```
gibizer, line 398: nit: galera/mariadb is now just galera

gibizer (approved):
Looks good to me.

controllers/mariadbaccount_controller.go
gibizer: I wanted to add this to L397 instead of L398 but github does not allow it

zzzeek:
waiting on @dciabrin  to /approve

PR #179, opened 2024-01-11
Remove secret from MariaDBDatabase and use MariaDBAccount instead

zzzeek:
this change proposes an initial step to **transition** to the use of MariaDBAccount to manage DB username/password combinations, rather than using MariaDBDatabase as is the case right now.

Downstream services make use of the `Database{}` API as an abstraction layer on top of `MariaDBDatabase` in order to ensure their database exists.    They call upon 1. `mariadbv1.NewDatabase` to create the object and then 2. `CreateOrPatchDB` to perform the work to make this database available.    

**Prior to this PR**, `CreateOrPatchDB` creates a `MariaDBDatabase` CR; `MariaDBDatabase` does both CREATE DATABASE as well as CREATE USER.   `MariaDBDatabase` includes a `secret` parameter that is combined with the name of the database itself as a hardcoded username to create a database plus a login.   the login cannot be changed independently of the database itself.   Even though `Database` has a separate username parameter `Database.databaseUser`, this parameter is currently ignored.

**The change made to this PR** is that `CreateOrPatchDB` creates a `MariaDBDatabase` CR **and** a `MariaDBAccount` CR, making use of the `Database.databaseUser` parameter to create both a database and a username/password that is not hardcoded and can vary independently of the database.   Appropriate finalizers are applied to both the `MariaDBDatabase` and `MariaDBAccount` CRs from the downstream controller.     

Towards this change, the `secret` parameter is **deprecated** in`MariaDBDatabase`, defaulting to non-present; `MariaDBDatabase` no longer creates a username/password by itself in the new database unless "secret" is present.   By doing so, `MariaDBAccount` is free to create database accounts where the username matches the name of the database, which is currently required by downstream controllers that all establish `Database.databaseName` and`Database.databaseUser` using the same name.        

The "secret" parameter continues to be supported by `MariaDBDatabase` in an optional way so that downstream operators built on previous versions of the mariadb-operator will still continue to function, since they are still generating `MariaDBDatabase` objects with "secrets" and they are not generating `MariaDBAccount` objects.  Once all downstream operators are built against this version of the code, where the `Database` construct creates a separate `MariaDBAccount`, then the "secret" parameter can be fully removed from `MariaDBDatabase` and the logic to generate and drop mariadb accounts can be removed from `database.go`.

These steps are intended only to integrate `MariaDBAccount` and independent username/passwords into the existing flow, without changing the flow itself, as this would be a breaking change in downstream controllers.  The downstream controllers right now pass an explicit "databaseuser" and "secret" into `Database`, not the other way around.  There is also no mechanism for downstream controllers to respond to changes in database username / password.   this can't be implemented until `MariaDBAccount` is integrated as the controllers would need to be watching and consuming their corresponding `MariaDBAccount` objects directly.

Introduction of an independent "account / password generation service" would be done in a subsequent step.  Subsequent steps would be:

1. downstream operators will build against this version of mariadb-operator or greater.
2. "secret" logic is removed from `MariaDBDatabase`.
3. downstream operators would need to be modified to become responsive to changes in MariaDBAccount CRs so that they can rewrite their configs and re-deploy pods when the MariaDBAccount they rely upon is to be replaced with a new one.
4. a new flow would need to be introduced that can generate as well as rotate username/passwords which can be deployed using the `MariaDBAccount` CR.   The nature of this flow is TBD.
5. downstream operators would no longer have any kind of "database username" or "database secret" in their own CRs - this would be created dynamically by an account rotation service.


api/v1beta1/mariadbdatabase_funcs.go
```diff
@@ -157,11 +162,28 @@ func (d *Database) CreateOrPatchDBByName(
 			},
 			Spec: MariaDBDatabaseSpec{
 				// the DB name must not change, therefore specify it outside the mutuate function
-				Name: d.databaseName,
+				Name:                d.databaseName,
+				DefaultCharacterSet: "utf8",
```
zzzeek, line 166: suggestions welcome how to set these defaults for the MariaDBDatabaseSpec

api/v1beta1/mariadbdatabase_funcs.go
```diff
@@ -287,6 +363,35 @@ func (d *Database) getDBWithName(
 
 	d.database = db
 
+	account := &MariaDBAccount{}
+	username := d.databaseUser
```
zzzeek, line 367: note that I'm not doing the "GetBeforeObject()" thing here, that seems to be done with MariaDBDatabase name/namespace; the object we get with GetBeforeObject doesn't actually have this "databaseUser" information.

Overall  I dont really understand what use case GetBeforeObject() is serving and I wonder if it's not needed anymore here, since my new additions seem to work with an actual consuming operator without "d.databaseUser" being blank.

gibizer: the main reason we have GetBeforeObject() in the helper is to be able to generate a Patch for the CR at the end of the reconcile run in https://github.com/openstack-k8s-operators/lib-common/blob/1715079755dc337f014a77561c6a665fa80ff3cd/modules/common/helper/helper.go#L193 
In general I would avoid depending on the helper in every new change if possible as it is a bag of unrelated things so depending on the helper introduce new unwanted dependencies as well.

api/v1beta1/mariadbdatabase_types.go
```diff
@@ -44,6 +43,9 @@ type MariaDBDatabaseSpec struct {
 
 // MariaDBDatabaseStatus defines the observed state of MariaDBDatabase
 type MariaDBDatabaseStatus struct {
+	// Deployment Conditions
+	Conditions condition.Conditions `json:"conditions,omitempty" optional:"true"`
```
gibizer, line 47: +1 for adding conditions.

controllers/mariadbdatabase_controller.go
```diff
@@ -161,6 +197,11 @@ func (r *MariaDBDatabaseReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		serviceAccount = dbMariadb.RbacResourceName()
 	}
 
+	instance.Status.Conditions.MarkTrue(
```
gibizer, line 200: You can set the MariaDBServerReadyCondition to False with a message at L180 and L190 to indicate that the DB bootstrap is still ongoing. So the CR status will show the progress.

gibizer: Also is case of error from `r.getDatabaseObject()` you can report that error in this condition too

controllers/mariadbdatabase_controller.go
```diff
@@ -193,8 +234,10 @@ func (r *MariaDBDatabaseReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		log.Info("Job hash added", "Job", jobDef.Name, "Hash", instance.Status.Hash[databasev1beta1.DbCreateHash])
 	}
 
-	// database creation finished... okay to set to completed
-	instance.Status.Completed = true
+	instance.Status.Conditions.MarkTrue(
```
gibizer, line 237: ditto you can report errors or in progress work related to the Job in this condition above

controllers/mariadbdatabase_controller.go
```diff
@@ -74,13 +75,48 @@ func (r *MariaDBDatabaseReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 
 	// Always patch the instance status when exiting this function so we can persist any changes.
 	defer func() {
+		if instance.Status.Conditions.AllSubConditionIsTrue() {
```
gibizer, line 78: +1 thanks for following the pattern

api/v1beta1/mariadbdatabase_funcs.go
```diff
@@ -157,11 +162,28 @@ func (d *Database) CreateOrPatchDBByName(
...
-				Name: d.databaseName,
+				Name:                d.databaseName,
+				DefaultCharacterSet: "utf8",
+				DefaultCollation:    "utf8_general_ci",
 			},
 		}
 	}
 
+	account := d.account
+	if account == nil {
```
gibizer, line 173: I guess it is here as a stop gap until the service operators are updated to pass in the Account via  NewDatabase* calls. Based on the spec in the google doc we expect that the account is passed in not created within the mariadb-operator.

api/v1beta1/mariadbdatabase_funcs.go
```diff
@@ -157,11 +162,28 @@ func (d *Database) CreateOrPatchDBByName(
...
+		account = &MariaDBAccount{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      d.databaseUser,
+				Namespace: d.namespace,
+				Labels: map[string]string{
+					"mariaDBDatabaseName": d.name,
+				},
+			},
+			Spec: MariaDBAccountSpec{
+				UserName: d.databaseUser,
```
gibizer, line 183: I guess this will be removed later and the user name is read from the passed in account instead.

api/v1beta1/mariadbdatabase_funcs.go
gibizer: Unfortunately golang and json defaulting is not working nicely together. When you instantiate MariaDBDatabaseSpec struct here, golang will fill all the undefined fields of the struct with the golang default value of the field type. For string it is `""`. Then when you send this struct to the json serializer (when send it to the k8s API) the json serializer sees the `""` value and puts that into the json payload. Then when that is deserialized by the API from json to golang, it will see that the field is specified so no defaulting is applied.

You can work around this by telling the json serializer not to emit the field to the json payload if the value of the field is the golang default value of the field type. The keyword is `omitempty`

The following def will consider `""` as a value that means no value is provided:
```golang
	// +kubebuilder:default=utf8
	// Default character set for this database
	DefaultCharacterSet string `json:"defaultCharacterSet,omitempty"`
```

If the empty string `""` is a valid field value for the CR then you can do an additional step to keep `""` valid. You can make the field as a string pointer in golang. This way `""` is a valid value and `nil` will be used to signal unspecified value instead:
```golang
	// +kubebuilder:default=utf8
	// Default character set for this database
	DefaultCharacterSet *string `json:"defaultCharacterSet,omitempty"`
```

api/v1beta1/mariadbdatabase_funcs.go
```diff
@@ -200,6 +220,38 @@ func (d *Database) CreateOrPatchDBByName(
 		return ctrl.Result{RequeueAfter: time.Second * 5}, nil
 	}
 
+	op_acc, err_acc := controllerutil.CreateOrPatch(ctx, h.GetClient(), account, func() error {
+		account.Labels = util.MergeStringMaps(
+			account.GetLabels(),
+			d.labels,
+		)
+
+		account.Spec.Secret = d.secret
```
gibizer, line 229: This seems strange. According to the spec in the google doc we want to read the secret from the account not vice versa. I expect that the service operators are passing a fully populated account to mariadb-operator and the operator will use that to create the user/password.

> The credentials generator role will apply the name of this CR to the consumer operator’s CR ; e.g. the Neutron CR will refer to a new GaleraAccount CR that it should use for database credentials

zzzeek: it does seem strange, I was imitating how the previous MariaDBDatabase code was doing it though I didnt understand yet why it was that way:  

https://github.com/openstack-k8s-operators/mariadb-operator/blob/fe72003c634354c9ba13fd7d330f83e03119fc7b/api/v1beta1/mariadbdatabase_funcs.go#L171-L179

api/v1beta1/mariadbdatabase_funcs.go
zzzeek: OK, so ill leave this as it is for now....

api/v1beta1/mariadbdatabase_funcs.go
zzzeek: mmmm not really sure, the whole point of MariaDBAccount is that the usernames *also* change, so are independent of the database name.   the usernames change so that we can do password rotations with no downtime.

api/v1beta1/mariadbdatabase_funcs.go
zzzeek: > This seems strange. According to the spec in the google doc we want to read the secret from the account not vice versa. I expect that the service operators are passing a fully populated account to mariadb-operator and the operator will use that to create the user/password.

for now, just to decouple MariaDBDatabase from username without breaking all the downstream services, the idea is to keep the Database{} object exactly the same, which currently includes a database username that is passed in from controllers like that of glance and nova (I checked), and to create a MariaDBAccount when they do so.

> > The credentials generator role will apply the name of this CR to the consumer operator’s CR ; e.g. the Neutron CR will refer to a new GaleraAccount CR that it should use for database credentials

So we dont have as yet a "credentials generator" and my understanding is that we dont want to home-roll this, otherwise I'd just generate a password here.    So in the interests of moving forward to have MariaDBAccount participating in OSP 18 ahead of individual services being aware of it, right now the flow of Database->MariaDBAccount mirrors that of Database->MariaDBDatabase.

controllers/mariadbdatabase_controller.go
zzzeek: So that should happen in mariadbaccount controller also, which has the same code.  right now the logic is that it shows "unknown".

zzzeek: I added this to all four spots (database, account, both for galera and mariadb )

api/v1beta1/mariadbdatabase_funcs.go
zzzeek: OK it looks like adding omitempty worked so that I can remove those additional hardcoded charset/collation defaults from the go code

controllers/mariadbdatabase_controller.go
zzzeek: which is now two spots now that mariadb CR removal has landed

api/v1beta1/mariadbdatabase_funcs.go
```diff
@@ -200,6 +219,36 @@ func (d *Database) CreateOrPatchDBByName(
 		return ctrl.Result{RequeueAfter: time.Second * 5}, nil
 	}
 
+	op_acc, err_acc := controllerutil.CreateOrPatch(ctx, h.GetClient(), account, func() error {
+		account.Labels = util.MergeStringMaps(
+			account.GetLabels(),
+			d.labels,
+		)
+
+		err := controllerutil.SetControllerReference(h.GetBeforeObject(), account, h.GetScheme())
```
SeanMooney, line 228: what is the reason for h.GetBeforeObject() here.

in the nova operator we almost never use this function and its littered through out the mariadb operator code

for nova we do 

err := controllerutil.SetControllerReference(instance, novncproxy, r.Scheme)
		if err != nil {
			return err
		}

https://github.com/openstack-k8s-operators/nova-operator/blob/8f204945b887a562229143f40901c4540925e9cd/controllers/novacell_controller.go#L454

should we not be passing the database here or similar?

zzzeek: so this is the same as what I see on line 179 for MariaDBDatabase

the use of GetBeforeObject() is local to mariadbdatabase_funcs.go where these functions were originally in openstack-operator, see 6644f41727771341c6c18c

@gibizer can you recommend how this should be working and should we be removing the use of getBeforeObject() from this file?

SeanMooney: i chatted to gibi about this. tldr what we should be doing is passing the instance point into this function for the CR that is being modified instead of looking it up via the helper.


the helper has a copy of the original object before we reconsile it so that it can calculate the deleta when we patch the object at the end of the reconsile loop.

that's why in the nova operator we almost never use getBeforeObject
we pass the instance fo the CR object to the functions that need it directly i.e. we take a nova-api object pointer in the function signigure and pass that.

I'm not 100% sure but you may be able to just use d here as its an instance of the MariaDBDatabase object that you are reconsiling.

i have not checked if SetControllerReference is mutating the object that is passed to it but i would assume it is in which case passing the before object pointer is likely a bug
since it will break the ability to calculate the patch delta since you are modifying the before object

gibizer:
Excellent PR summary message @zzzeek . It explained lot of my questions about the intentions of this and the next steps.

api/v1beta1/mariadbdatabase_funcs.go
gibizer: The PR message explains this well. Thank.

api/v1beta1/mariadbdatabase_funcs.go
gibizer: You are right we need independently changeable user names.

api/v1beta1/mariadbdatabase_funcs.go
gibizer: > for now, just to decouple MariaDBDatabase from username without breaking all the downstream services, the idea is to keep the Database{} object exactly the same, which currently includes a database username that is passed in from controllers like that of glance and nova (I checked), and to create a MariaDBAccount when they do so.

Yeah, it make sense now.

api/v1beta1/mariadbdatabase_funcs.go
gibizer: So the intention here is to have a controller reference between the service CR (like PlacementAPI) and the new MariaDBAccount CR created for the service. This can only be done if this code has access to the service CR as it needs the kind/name/uid of the owner to set it on the owned. 

I think we can keep this as is for now. As far as I understand the MariaDBAccount creation here is temporary anyhow and we expect it to be moved to the service operator or even further up the stack. During such move the controller ref creation can be moved to the service operator and then the GetBeforeObject can be removed.

In my eyes having Helper is an anti pattern, it contains to many unrelated things and creates unfortunate dependencies. But breaking up Helper is a work for the future (if ever).

SeanMooney: i would advocate for the MariaDBAccount creation to move to the openstack-oeprator like we do with the galara dbs and rabbit. then just pass the name of the account crs to the service operators. that way rotation can be done by creating new account in anddtion to the current ones. updating the service cr references. then when its reconosled drop the old accounts. there would need to be an edpm deployment too but ya doing this all form the top-level sound like the right approach

gibizer: I think this is where we are headed. My point is if the MariaDBAccount is passed to the service operator eventually then the control reference on the MariaDBAccount can be added in the service operator code, and there we don't need the helper to access the service CR.

controllers/mariadbdatabase_controller.go
zzzeek: OK....I have a bunch in there, we can add more progress features as needed...

api/v1beta1/mariadbdatabase_funcs.go
zzzeek: OK so for now I think it's fine to leave what we have and can be refined in subsequent prs since we will be changing this flow in any case

zzzeek:
I just realized why this isn't building.   will have to pull back the changes to mariadbdatabase and have it work "both" ways for cross-compatibility

gibizer:
> I had to re-introduce "secret" to `MariaDBDatabase` since in the build, the downstream operators still build against the older version of the operator. this complicates the `MariaDBDatabase` operator since "secret" is now optional on that end.

It is fine by me. 
I think we are close to inevitable point of breaking the service operators and just go and fix them (I know it is a lot of work)

gibizer (approved):

PR #180, opened 2024-01-16
detect ipv6 stack and apply correct listen address

zzzeek:
on a dual-stack host (or an ipv6 only host), detect the ipv6 and ipv4 addresses individually from /etc/hosts and apply '[::]' as the gmcast_listen address if ipv6 is present.

Fixes: https://issues.redhat.com/browse/OSPRH-3330

### DIFF
--- a/templates/galera/bin/mysql_bootstrap.sh
+++ b/templates/galera/bin/mysql_bootstrap.sh
@@ -19,11 +19,22 @@ fi
 # Generate the mariadb configs from the templates, these will get
 # copied by `kolla_start` when the pod's main container will start
 PODNAME=$(hostname -f | cut -d. -f1,2)
-PODIP=$(grep "${PODNAME}" /etc/hosts | cut -d$'\t' -f1)
+PODIPV4=$(grep "${PODNAME}" /etc/hosts | grep -v ':' | cut -d$'\t' -f1)
+PODIPV6=$(grep "${PODNAME}" /etc/hosts | grep ':' | cut -d$'\t' -f1)
+
 cd /var/lib/config-data
 for cfg in *.cnf.in; do
     if [ -s "${cfg}" ]; then
-        echo "Generating config file from template ${cfg}"
+
+        if [[ "" = "${PODIPV6}" ]]; then
+            PODIP="${PODIPV4}"
+            IPSTACK="IPV4"
+        else
+            PODIP="[::]"
+            IPSTACK="IPV6"
+        fi
+
+        echo "Generating config file from template ${cfg}, will use ${IPSTACK} listen address of ${PODIP}"
         sed -e "s/{ PODNAME }/${PODNAME}/" -e "s/{ PODIP }/${PODIP}/" "/var/lib/config-data/${cfg}" > "/var/lib/pod-config-data/${cfg%.in}"
     fi
 done


hjensas:

`error: build error: Failed to push image: trying to reuse blob sha256:07a64a71e01156f8f99039bc246149925c6d1480d3957de78510bbec6ec68f7a at destination: pinging container registry quay.rdoproject.org: Get "https://quay.rdoproject.org/v2/": net/http: TLS handshake timeout`

templates/galera/bin/mysql_bootstrap.sh
```diff
@@ -19,11 +19,22 @@ fi
...
 cd /var/lib/config-data
 for cfg in *.cnf.in; do
     if [ -s "${cfg}" ]; then
-        echo "Generating config file from template ${cfg}"
+
+        if [[ "" = "${PODIPV6}" ]]; then
+            PODIP="${PODIPV4}"
+            IPSTACK="IPV4"
+        else
+            PODIP="[::]"
```
hjensas, line 33: Do we need all this logic?
 If we are defaulting to listen on all interfaces in case IPv6 or Dual-Stack IPv6-and-Ipv6 - Can we now simply configure it with "[::]" for all cases?

zzzeek: it has to work on an ipv4-only stack as well, IIUC galera needs the specific IP number to bind onto in that case cc @dciabrin

hjensas:

`error: build error: Failed to push image: trying to reuse blob sha256:07a64a71e01156f8f99039bc246149925c6d1480d3957de78510bbec6ec68f7a at destination: pinging container registry quay.rdoproject.org: Get "https://quay.rdoproject.org/v2/": net/http: TLS handshake timeout`

zzzeek:
now it passes after i put a node hold in.  faceslap

hjensas (approved):
I'm ok with this.
But I do find it weird that we choose to specify a "listen" address to bind to when using IPv4 - but then if it is IPv6 or IPv4+IPv6 (dual stack) we accept a config that make it listen to any interface.

It seems to me we can use either '0.0.0.0' or '[::]' for all cases instead.

zzzeek:
> I'm ok with this. But I do find it weird that we choose to specify a "listen" address to bind to when using IPv4 - but then if it is IPv6 or IPv4+IPv6 (dual stack) we accept a config that make it listen to any interface.
> 
> It seems to me we can use either '0.0.0.0' or '[::]' for all cases instead.

@dciabrin can you comment on this?

dciabrin:
> > I'm ok with this. But I do find it weird that we choose to specify a "listen" address to bind to when using IPv4 - but then if it is IPv6 or IPv4+IPv6 (dual stack) we accept a config that make it listen to any interface.
> > It seems to me we can use either '0.0.0.0' or '[::]' for all cases instead.
> 
> @dciabrin can you comment on this?

I'm still unclear about the general consensus as to listen in container: should we explicitly listen on the interface we expect to receive traffic from, or is listening on all interface and assume k8s plumbing is sufficient to not expose unnecessary interfaces.  As I said earlier this IPv6 patch is what we used back in the days because we had issues specifying IPv6 addresses with galera.

So I think this is good enough to fix the original reported issue, and we should probably revisit it to either specify a full IPv6 address is recent galera allows it, or listen to all interfaces if that is a  general pattern in pods networking.

templates/galera/bin/mysql_bootstrap.sh
dciabrin: We started using [::] for IPv6 as per https://github.com/openstack/puppet-tripleo/commit/84ba7c31fb6ed734f7c272cdb1e157e6cbecde98.
Galera is probably able to cope with other direct IPv6 addresses now, but that is to be verified. Until then, I think we are good enough with that as it's not a regression from what we used to have.

dciabrin (approved):

PR #183, opened 2024-01-29
when generating metadata.name, use dashes instead of underscores

zzzeek:
The Database{} facade that is now used to generate a MariaDBDatabase as well as a MariaDBAccount is currently generating a metadata name for the MariaDBAccount based on the username.  As we do in other places, if this username has underscores in it as is common, swap them out with dashes.


gibizer (approved):
I'm OK with this as a stop-gap. The intention is to have the user name generated so eventually we can make sure that no two user names will be generated that result in the same MariaDBAccount name.

PR #184, opened 2024-01-29
add functions to support downstream generation of MariaDBAccounts by operators

zzzeek:
these changes are to support a phased approach to getting openstack-operator to generate MariaDBAccounts.   I have an initial demo for glance that uses these functions

the glance PR is at https://github.com/openstack-k8s-operators/glance-operator/pull/426/

### DIFF
--- a/api/test/helpers/harnesses.go
+++ b/api/test/helpers/harnesses.go
@@ -1,0 +1,480 @@
+/*
+Copyright 2023 Red Hat
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package helpers
+
+import (
+	"fmt"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	mariadbv1 "github.com/openstack-k8s-operators/mariadb-operator/api/v1beta1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// populateHarness describes a function that will insert suite-appropriate
+// data into a MariaDBTestHarness instance
+type populateHarness func(*MariaDBTestHarness)
+
+// establishesCR describes a test function that can fully set up a particular
+// controller's "Reconciliation Successful" state for a given kind of CR.
+type establishesCR func(types.NamespacedName)
+
+// updatesAccountName describes a test function that can change the
+// "databaseAccount" or similar member of an already-reconciled CR to a new
+// one, which is expected to kick off a username/password rotation sequence.
+type updatesAccountName func(types.NamespacedName)
+
+// deletesCr describes a test function that will delete the CR that was
+// created by an establishesCR function
+type deletesCR func()
+
+type assertsURL func(types.NamespacedName, string, string)
+
+type getsConfigHash func() string
+
+// MariaDBTestHarness describes the parameters for running a series
+// of Ginkgo tests which exercise a controller's ability to correctly
+// work with MariaDBDatabase / MariaDBAccount APIs.
+type MariaDBTestHarness struct {
+	description     string
+	namespace       string
+	databaseName    string
+	finalizerName   string
+	PopulateHarness populateHarness
+	SetupCR         establishesCR
+	UpdateAccount   updatesAccountName
+	DeleteCR        deletesCR
+	mariaDBHelper   *TestHelper
+	timeout         time.Duration
+	interval        time.Duration
+}
+
+func (harness *MariaDBTestHarness) Setup(
+	description,
+	namespace string,
+	databaseName string,
+	finalizerName string,
+	mariadb *TestHelper,
+	timeout time.Duration,
+	interval time.Duration,
+) {
+	harness.description = description
+	harness.namespace = namespace
+	harness.databaseName = databaseName
+	harness.finalizerName = finalizerName
+	harness.mariaDBHelper = mariadb
+	harness.timeout = timeout
+	harness.interval = interval
+}
+
+// RunBasicSuite runs MariaDBAccount suite tests.  these are
+// pre-packaged ginkgo tests that exercise standard account create / update
+// patterns that should be common to all controllers that work with
+// MariaDBDatabase and MariaDBAccount CRs.
+func (harness *MariaDBTestHarness) RunBasicSuite() {
+
+	When(fmt.Sprintf("The %s service is being configured to run", harness.description), func() {
+		BeforeEach(func() {
+			harness.init()
+		})
+
+		It("Uses a pre-existing MariaDBAccount and sets a finalizer", func() {
+
+			mariaDBHelper, timeout, interval := harness.mariaDBHelper, harness.timeout, harness.interval
+
+			k8sClient := mariaDBHelper.K8sClient
+
+			accountName := types.NamespacedName{
+				Name:      "some-mariadb-account",
+				Namespace: harness.namespace,
+			}
+
+			// create MariaDBAccount first
+			acc, accSecret := mariaDBHelper.CreateMariaDBAccountAndSecret(accountName, mariadbv1.MariaDBAccountSpec{})
+			DeferCleanup(k8sClient.Delete, mariaDBHelper.Ctx, accSecret)
+			DeferCleanup(k8sClient.Delete, mariaDBHelper.Ctx, acc)
+
+			// then create the CR
+			harness.SetupCR(accountName)
+
+			mariaDBHelper.Logger.Info(fmt.Sprintf("Service should fully configure on MariaDBAccount %s", accountName))
+
+			// now wait for the account to exist
+			mariadbAccount := mariaDBHelper.GetMariaDBAccount(accountName)
+			Expect(mariadbAccount.Spec.UserName).ShouldNot(Equal(""))
+			Expect(mariadbAccount.Spec.Secret).ShouldNot(Equal(""))
+			mariaDBSecret := harness.mariaDBHelper.GetSecret(types.NamespacedName{Name: mariadbAccount.Spec.Secret, Namespace: mariadbAccount.Namespace})
+			Expect(string(mariaDBSecret.Data[mariadbv1.DatabasePasswordSelector])).ShouldNot(Equal(""))
+
+			// wait for finalizer to be present
+			Eventually(func() []string {
+				mariadbAccount := mariaDBHelper.GetMariaDBAccount(accountName)
+				return mariadbAccount.Finalizers
+			}, timeout, interval).Should(ContainElement(harness.finalizerName))
+
+			// mariaDBDatabaseName is set
+			Expect(mariadbAccount.Labels["mariaDBDatabaseName"]).Should(Equal(harness.databaseName))
+
+		})
+
+		It("Ensures a MariaDBAccount is created if not present and sets a finalizer", func() {
+			mariaDBHelper, timeout, interval := harness.mariaDBHelper, harness.timeout, harness.interval
+
+			accountName := types.NamespacedName{
+				Name:      "some-mariadb-account",
+				Namespace: harness.namespace,
+			}
+
+			// here, dont create a mariadbaccount.  right now CRs should
+			// generate this if not exists using EnsureMariaDBAccount
+
+			// then create the CR
+			harness.SetupCR(accountName)
+
+			mariaDBHelper.Logger.Info(fmt.Sprintf("Service should fully configure on MariaDBAccount %s", accountName))
+
+			// now wait for the account to have the finalizer and the
+			// database name
+			// now wait for the account to exist
+			mariadbAccount := mariaDBHelper.GetMariaDBAccount(accountName)
+			Expect(mariadbAccount.Spec.UserName).ShouldNot(Equal(""))
+			Expect(mariadbAccount.Spec.Secret).ShouldNot(Equal(""))
+			mariaDBSecret := harness.mariaDBHelper.GetSecret(types.NamespacedName{Name: mariadbAccount.Spec.Secret, Namespace: mariadbAccount.Namespace})
+			Expect(string(mariaDBSecret.Data[mariadbv1.DatabasePasswordSelector])).ShouldNot(Equal(""))
+
+			// wait for finalizer to be present
+			Eventually(func() []string {
+				mariadbAccount := mariaDBHelper.GetMariaDBAccount(accountName)
+				return mariadbAccount.Finalizers
+			}, timeout, interval).Should(ContainElement(harness.finalizerName))
+
+			// mariaDBDatabaseName is set
+			Expect(mariadbAccount.Labels["mariaDBDatabaseName"]).Should(Equal(harness.databaseName))
+
+		})
+	})
+
+	When(fmt.Sprintf("The %s service is fully running", harness.description), func() {
+		BeforeEach(func() {
+			harness.init()
+		})
+
+		// get service fully complete with a mariadbaccount
+		BeforeEach(func() {
+			mariaDBHelper, timeout, interval := harness.mariaDBHelper, harness.timeout, harness.interval
+
+			oldAccountName := types.NamespacedName{
+				Name:      "some-old-account",
+				Namespace: harness.namespace,
+			}
+
+			// create the CR with old account
+			harness.SetupCR(oldAccountName)
+
+			// also simulate that it got completed
+			mariaDBHelper.SimulateMariaDBAccountCompleted(oldAccountName)
+
+			mariaDBHelper.Logger.Info(fmt.Sprintf("Service should fully configure on MariaDBAccount %s", oldAccountName))
+
+			// finalizer is attached to old account
+			Eventually(func() []string {
+				oldMariadbAccount := mariaDBHelper.GetMariaDBAccount(oldAccountName)
+				return oldMariadbAccount.Finalizers
+			}, timeout, interval).Should(ContainElement(harness.finalizerName))
+
+		})
+		It("should ensure a new MariaDBAccount exists when accountname is changed", func() {
+			mariaDBHelper, timeout, interval := harness.mariaDBHelper, harness.timeout, harness.interval
+
+			oldAccountName := types.NamespacedName{
+				Name:      "some-old-account",
+				Namespace: harness.namespace,
+			}
+
+			newAccountName := types.NamespacedName{
+				Name:      "some-new-account",
+				Namespace: harness.namespace,
+			}
+
+			mariaDBHelper.Logger.Info("About to update account from some-old-account to some-new-account")
+
+			harness.UpdateAccount(newAccountName)
+
+			// new account is (eventually) created
+			_ = mariaDBHelper.GetMariaDBAccount(newAccountName)
+
+			// dont simuluate MariaDBAccount being created. it's not done yet
+
+			mariaDBHelper.Logger.Info(
+				fmt.Sprintf("Service should have ensured MariaDBAccount %s exists but should remain running on %s",
+					newAccountName, oldAccountName),
+			)
+
+			// finalizer is attached to new account
+			Eventually(func() []string {
+				newMariadbAccount := mariaDBHelper.GetMariaDBAccount(newAccountName)
+				return newMariadbAccount.Finalizers
+			}, timeout, interval).Should(ContainElement(harness.finalizerName))
+
+			// old account retains the finalizer because we did not yet
+			// complete the new MariaDBAccount
+			Consistently(func() []string {
+				oldMariadbAccount := mariaDBHelper.GetMariaDBAccount(oldAccountName)
+				return oldMariadbAccount.Finalizers
+			}, timeout, interval).Should(ContainElement(harness.finalizerName))
+		})
+
+		It("should move the finalizer to a new MariaDBAccount when create is complete", func() {
+			mariaDBHelper, timeout, interval := harness.mariaDBHelper, harness.timeout, harness.interval
+
+			oldAccountName := types.NamespacedName{
+				Name:      "some-old-account",
+				Namespace: harness.namespace,
+			}
+
+			newAccountName := types.NamespacedName{
+				Name:      "some-new-account",
+				Namespace: harness.namespace,
+			}
+
+			harness.UpdateAccount(newAccountName)
+
+			harness.runAccountUpdateWithWait(oldAccountName, newAccountName)
+
+			// CreateOrPatchDBByName will add a label referring to the database
+			Eventually(func() string {
+				mariadbAccount := mariaDBHelper.GetMariaDBAccount(newAccountName)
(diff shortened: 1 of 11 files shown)


api/v1beta1/mariadbdatabase_funcs.go
```diff
@@ -459,3 +517,127 @@ func (d *Database) DeleteFinalizer(
...
+}
+
+// InterimCreateNewAccount generates a new MariaDBAccount with a generated
+// password, if an account of that name does not yet exist.
+// this function is not intended to live long term within
+// the mariadb operator and is instead an interim step for individual
+// downstream operators to generate a new account, which would then be taken
+// over by the openstack-operator once each downstream is using the new
+// flow.
+func InterimCreateNewAccount(ctx context.Context,
```
zzzeek, line 573: i named this "Interim" because I really dont want people getting scared that I am adding account / pw generation into the mariadb operator, as that idea got a lot of pushback when I last suggested it.  it's here for now because it has to be *somewhere*, and we can move this function, or the "generate the password" part of it, to any operator or utility later.   

that said, naming this "interim" might not be a great idea as we can all see in two years everyone is still calling "InterimCreateNewAccount" and people are like "why is it 'interim'?"  :)

gibizer (commented):
looks good to me

api/v1beta1/mariadbdatabase_funcs.go
```diff
@@ -459,3 +517,127 @@ func (d *Database) DeleteFinalizer(
...
+		account = &MariaDBAccount{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      accountName,
+				Namespace: namespace,
+				// note no labels yet; the account will not have a
+				// mariadbdatabase yet so the controller will not
+				// try to create a DB; it instead will respond again to the
+				// MariaDBAccount once this is filled in
+			},
+			Spec: MariaDBAccountSpec{
```
olliewalsh, line 619: Should this manage RequireTLS?

zzzeek: yes that should be passed here,  which means we'd add it to `Database`.   However, it seems like we'd have to get to the point that openstack-operator is creating the MariaDBAccounts and not the individual operators themselves, to have this controlled by a configuration option.

zzzeek: I think RequireTLS should not be part of `Database` here, as the "long term" (where long term == hopefully in the next 6 weeks) every operator will be using `EnsureMariaDBAccount` as the single point of creation for new MariaDBAccount objects, and that will later go up to openstack-operator.   the ad-hoc creation of MariaDBAccount within CreateOrPatchDBByName is also going to go away.   So requireTLS should only be passed to EnsureMariaDBAccount  IMO.

zzzeek: OK....thats where this is now

api/v1beta1/mariadbdatabase_types.go
```diff
@@ -85,15 +85,17 @@ const (
...
+// Database - a facade on top of the combination of a MariaDBDatabase
+// and MariaDBAccount pair
 type Database struct {
 	database         *MariaDBDatabase
 	account          *MariaDBAccount
 	databaseHostname string
 	databaseName     string
-	databaseUser     string
-	secret           string
+	databaseUser     string // legacy; will go away when operators fully integrate MariaDBAccount
```
zzzeek, line 95: note `Database` seems to be used for *sending* instructions to mariadb-operator but my early impression is that it isn't used for *consumption*, hence these can go away as controllers would consume from the mariadbaccount directly.     I lean towards that because I find the `Database` facade to be kind of awkward/overcomplicated and potentially error prone to the degree that it repeats information in local fields vs. referenced CRs since decisions have to be made when and how to move values between local fields vs. CRs.

however if people want to chime in that we do want `Database` to have local fields for username/password/requireTLS even and that these would be bidirectional, I'd be willing to hear the rationale for that.

cc @gibizer

gibizer: I agree that we should remove these fields and use the MariaDBAccount instead.

api/v1beta1/mariadbdatabase_funcs.go
```diff
@@ -457,3 +515,185 @@ func (d *Database) DeleteFinalizer(
 	}
 	return nil
 }
+
+// DeleteUnusedMariaDBAccountFinalizers searches for all MariaDBAccounts
+// associated with the given MariaDBDabase name and removes the finalizer for all
+// of them except for the given named account.
+func DeleteUnusedMariaDBAccountFinalizers(
```
gibizer, line 522: +1

api/v1beta1/mariadbdatabase_funcs.go
```diff
@@ -457,3 +515,185 @@ func (d *Database) DeleteFinalizer(
...
+	mariaDBAccountName string,
+	namespace string,
+) error {
+
+	accountList := &MariaDBAccountList{}
+
+	opts := []client.ListOption{
+		client.InNamespace(namespace),
+		client.MatchingLabels{"mariaDBDatabaseName": mariaDBDatabaseName},
+		client.MatchingFields{"kind": "MariaDBAccount"},
```
zzzeek, line 535: i dont know if this part of the criteria is needed or if it sees we passed a MariaDBAccountList already

api/v1beta1/mariadbdatabase_funcs.go
zzzeek: just ran a test and this function didn't actually work, the finalizer is still in the old mariadbaccount when i did an account switch in an openstack cluster.  everything else worked so ill do more testing of this next week.

api/v1beta1/mariadbdatabase_funcs.go
zzzeek: this has been renamed to be a general EnsureAccount function that can be referenced permanently.

api/v1beta1/mariadbdatabase_funcs.go
zzzeek: I was right about the "kind"  

      {"error": "Index with name field:kind does not exist"}

zzzeek: OK so i have this working, but now I have a doubt.   Why do we need a Glance finalizer in MariaDBAccount, if we are still deleting the actual MySQL database when the deletion timestamp on the CR is set, without us actually checking the finalizers?  I can mark a MariaDBAccount as deleted, mariadb-operator emits DROP USER, all while the glance-operator is still assuming that's the current account.

Should mariadb-operator be only emitting "DROP USER" once all finalizers other than the "self" finalizer is removed , that is, if there's any other finalizers in the MariaDBAccount such as "glance", dont do the DROP USER yet ?    do we have any examples of this flow elsewhere?

zzzeek: I think the answer is yes, and if it's not, we can revert the code I just added that does this.  I think this makes things a lot more robust since the mariadb user is not going to go away until the downstream controller's finalizer is removed.

api/v1beta1/mariadbdatabase_funcs.go
zzzeek: it's not valid, removed it and this works now.  also added logging for errors.

controllers/mariadbaccount_controller.go
```diff
@@ -389,6 +390,32 @@ func (r *MariaDBAccountReconciler) reconcileDelete(
 		return ctrl.Result{}, err
 	}
 
+	// dont do actual DROP USER until finalizers from downstream controllers
```
zzzeek, line 393: new logic to check for finalizers outside of "MariaDBAccount"

api/v1beta1/mariadbdatabase_funcs.go
```diff
@@ -457,3 +515,203 @@ func (d *Database) DeleteFinalizer(
...
+				Name:      accountName,
+				Namespace: namespace,
+				// note no labels yet; the account will not have a
+				// mariadbdatabase yet so the controller will not
+				// try to create a DB; it instead will respond again to the
+				// MariaDBAccount once this is filled in
+			},
+			Spec: MariaDBAccountSpec{
+				UserName:   username,
+				Secret:     fmt.Sprintf("%s-db-secret", accountName),
```
zzzeek, line 642: use `<name>-db-secret` for secret names since these are otherwise not database specific  cc @fmount

api/v1beta1/mariadbdatabase_funcs.go
gibizer: Good point. I agree that we need to wait for the state when the only finalizer on the MariaDBAccount is its own, then do DROP USER, and then remove the last finalizer.

api/v1beta1/mariadbdatabase_funcs.go
fmount: Thank you, makes sense and would help on ci/support front when doing the must-gather!

zzzeek:
I think this is very close to something that can merge, we might want to change some names of things or otherwise move the internal API around, though I think it's most important to just get everyone onto a new scheme that consumes username/password from MariaDBAccount.

So the Q here is now.  How can we confirm that merging this will **not** break the existing operators as they slowly start to build against it, some do , some dont, etc. ?     We just had a huge pass through all the operators to get them to look for MariaDBAccount in their functional test suites, that's great.  However it seemed like my pushing some of the MariaDBAccount changes forced the issue and forced something of a firedrill to happen.  I dont want that here.  So what do we need to do to ensure that?

zzzeek:
this now includes a test harness which exercises that a given controller does the following:

1. uses an existing MariaDBAccount when the service is created, adds finalizer and establishes DB with this username
2. if no exisitng MariaDBAccount, creates one using EnsureMariaDBAccount
3. if databaseAccount in the CR is changed, the controller ensures that new account exists, updates finalizers, and only upon successful creation for that account, removes finalizers from the old account

this suite now works for nova (api DB and cell DB), neutron, keystone, glance

given current deadlines we probably should look to merge this ASAP

zzzeek:
im making a note here for myself

checklist for consuming CRs:

1. controller calls ensuremariadbaccount
2. error code from ensuremariadbaccount is handled, Conditions are amended when error is returned
3. controller calls NewDatabaseForAccount instead of NewDatabase
4. GetAccountAndSecret is used to retrieve account /secret to populate template
5. GetDatabaseByName() , normally used for delete finalizers, replaced with GetDatabaseByNameAndAccount
6. controller calls DeleteUnusedMariaDBAccountFinalizers, returns error code if present
7. PasswordSelectors that refer to database are removed
8. all databaseUser replaced with databaseAccount
9.

zzzeek (commented):
the API functions here need to be fixed up because they sort of dont make sense.  what is the "name" passed to CreateOrPatchDBByName and why is that separate from the "dbName" label , that does not seem to be used in the legacy version of the code?  why are there two ways of getting dbName?  Why if someone used NewDatabaseForAccount do they need to re-send the "name" in after that anyway?   let's fix all this up

api/v1beta1/mariadbdatabase_funcs.go
```diff
@@ -66,10 +75,30 @@ func NewDatabaseWithNamespace(
...
+	databaseInstanceName string,
+	databaseName string,
+	name string,
+	accountName string,
+	namespace string,
+) *Database {
+	return &Database{
+		databaseName: databaseName,
+		labels: map[string]string{
+			"dbName": databaseInstanceName,
```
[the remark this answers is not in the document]
zzzeek: OK the controllers when they are in legacy are all sending this label in manually, like this:

```
   db := mariadbv1.NewDatabase(
               instance.Name,
               instance.Spec.DatabaseUser,
               instance.Spec.Secret,
               map[string]string{
                       "dbName": instance.Spec.DatabaseInstance,
               },
```

that's where dbName comes from so it's why in our new function, we receive it as an argument and put it there.

zzzeek:
i've made updates to the mariadbfuncs so that they make more sense from the outside, deprecated elements are moved outwards to deprecated functions, and redundant variables are not passed in for state we already have.

given things are moving fast I am ready for this to merge, it is doing its job at the level of account create, account update, deletion of old accounts, and testing it all via ginkgo functions packaged here, I can continue the related patches based on this logic and get them also ready for merge as they wont need a go.mod replace anymore.

cc @dciabrin @gibizer @stuggi

zzzeek:
OK one adjustment is that based on looking at PRs like https://github.com/openstack-k8s-operators/heat-operator/pull/323 I might want to use the pattern I suggested at the end of my comment at https://github.com/openstack-k8s-operators/heat-operator/pull/323/files#r1500046605 .   Using that idea, I would remove the GetAccountAndSecret method here and go back to using GetDatabaseByNameAndAccount to get a Database facade , where we'd get both the TLS info from as well as the mariadbaccount and secret

api/v1beta1/mariadbdatabase_funcs.go
```diff
@@ -374,76 +490,96 @@ func (d *Database) getDBWithName(
...
-			types.NamespacedName{
-				Name:      strings.Replace(username, "_", "-", -1),
-				Namespace: namespace,
-			},
-			account)
+		accountName = d.name
+		legacyAccount = true
 	}
 
+	mariaDBAccount, secretObj, err := GetAccountAndSecret(ctx, h, accountName, namespace)
```
zzzeek, line 507: loading both MariaDBAccount and the Secret here.  to suit @stuggi 's TLS patches, there are some places where we want to get all three things at once to populate templates / configmaps, so let the Database facade suit that for us.    Other places we still want to use just GetAccountAndSecret.

api/v1beta1/mariadbdatabase_funcs.go
```diff
@@ -496,3 +648,254 @@ func (d *Database) GetDatabaseClientConfig(s *tls.Service) string {
...
+// front.
+func EnsureMariaDBAccount(ctx context.Context,
+	helper *helper.Helper,
+	accountName string, namespace string, requireTLS bool,
+	userNamePrefix string,
+) (*MariaDBAccount, *corev1.Secret, error) {
+
+	account, err := GetAccount(ctx, helper, accountName, namespace)
+
+	if accountName == "" {
```
gibizer, line 788: would it make more sense to do this check before the GetAccount call as it cannot succeed if no name was provided

api/v1beta1/mariadbdatabase_funcs.go
```diff
@@ -124,151 +168,220 @@ func (d *Database) GetDatabaseHostname() string {
...
 
-	account := d.account
-	if account == nil {
-		// no account is present in this Database, so for forwards compatibility,
-		// make one.  name it the same as the MariaDBDatabase so we can get it back
-		// again based on that name alone.
-		account = &MariaDBAccount{
+	mariaDBAccount := d.account
+
+	if mariaDBAccount == nil {
```
gibizer, line 291: am I correct that this can only happen in the legacy code path as the new one will call EnsureMariaDBAccount before calling CreateOrPatchAll, so the account exists and loaded into the facade? If so then lets add a comment here that this path can be removed after the transition PRs merged

api/v1beta1/mariadbdatabase_funcs.go
```diff
@@ -282,18 +395,15 @@ func (d *Database) WaitForDBCreatedWithTimeout(
 	requeueAfter time.Duration,
 ) (ctrl.Result, error) {
 
-	err := d.getDBWithName(
-		ctx,
-		h,
-	)
+	err := d.loadDatabaseAndAccountCRs(ctx, h)
```
gibizer, line 398: do we really need to load it again? I think both the old and the new codepath already loaded it with previous calls(e.g.: NewDatabaseForAccount and CreateOrPatchAll) on the db facade. Probably not an issue to load it again, so it is more like a question to validate my understanding, than a request for a change here.

api/v1beta1/mariadbdatabase_funcs.go
```diff
@@ -313,7 +423,11 @@ func (d *Database) WaitForDBCreatedWithTimeout(
 	if k8s_errors.IsNotFound(err) {
 		util.LogForObject(
 			h,
-			fmt.Sprintf("DB or account objects not yet found %s", d.database.Name),
+			fmt.Sprintf(
+				"MariaDBDatabase %s and/or MariaDBAccount %s not yet found",
+				d.database.Name,
+				d.accountName,
+			),
 			d.database,
```
gibizer, line 431: I would move this check up  before the database status condition checking as checking the status on something that does not exist yet feels bad

api/v1beta1/mariadbdatabase_funcs.go
stuggi: agreed, accountName and namespace must not be ""

api/test/helpers/harnesses.go
```diff
@@ -0,0 +1,330 @@
...
+				Namespace: harness.Namespace,
+			}
+
+			// create MariaDBAccount first
+			acc, accSecret := mariadb.CreateMariaDBAccountAndSecret(accountName, mariadbv1.MariaDBAccountSpec{})
+			DeferCleanup(k8sClient.Delete, mariadb.Ctx, accSecret)
+			DeferCleanup(k8sClient.Delete, mariadb.Ctx, acc)
+
+			// then create the CR
+			crSetup(accountName)
```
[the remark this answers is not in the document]
zzzeek: the assumption here is the given crSetup() looks like a BeforeEach from the consuming suite, where the defercleanups are embedded.  see for example https://github.com/openstack-k8s-operators/nova-operator/blob/b4c144a94c905c18fc38c0c6f4ced5c7c31d9824/test/functional/nova_controller_test.go#L985

api/test/helpers/harnesses.go
```diff
@@ -0,0 +1,330 @@
...
+				mariadbAccount := mariadb.GetMariaDBAccount(accountName)
+				return mariadbAccount.Finalizers
+			}, harness.Timeout, harness.Interval).Should(ContainElement(harness.FinalizerName))
+
+			// CreateOrPatchDBByName will add a label referring to the database
+			Eventually(func() string {
+				mariadbAccount := mariadb.GetMariaDBAccount(accountName)
+				return mariadbAccount.Labels["mariaDBDatabaseName"]
+			}, harness.Timeout, harness.Interval).Should(Equal(harness.DatabaseName))
+
```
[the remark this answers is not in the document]
zzzeek: sure that should be doable

api/test/helpers/harnesses.go
```diff
@@ -0,0 +1,330 @@
...
+			// finalizer is attached to new account
+			Eventually(func() []string {
+				newMariadbAccount := mariadb.GetMariaDBAccount(newAccountName)
+				return newMariadbAccount.Finalizers
+			}, harness.Timeout, harness.Interval).Should(ContainElement(harness.FinalizerName))
+
+			// old account retains the finalizer because we did not yet
+			// complete the new MariaDBAccount
+			Eventually(func() []string {
```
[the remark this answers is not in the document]
zzzeek: ah, well , TIL, let's do that!

api/v1beta1/mariadbdatabase_funcs.go
zzzeek: the nil part here is not legacy because this is given a Database struct with just names in it, this function does the actual filling in with a MariaDBDatabase and MariaDBAccount object based on those names.  so even if the MariaDBAccount already exists, this part is loading it in.   this is the same flow that was already there.   

the legacy part here is the `if d.accountName == "" {` part, that wont happen in code that uses the newer functions.

api/v1beta1/mariadbdatabase_funcs.go
zzzeek: well this code is still being sloshed around from what it was previously, and it was previously doing this part.   I guess what this does is add some robustness if someone did not call CreateOrPatchXYZ first.     

I think what we might want to do, now that the controllers are getting more tightly focused ensureDB functions, is make some facade that rolls in this whole dance into one function.  Like here is keystone right now (this is with the patch, but it's similar before my patch now that we have @stuggi 's TLS changes):

```
func (r *KeystoneAPIReconciler) ensureDB(
	ctx context.Context,
	h *helper.Helper,
	instance *keystonev1.KeystoneAPI,
) (*mariadbv1.Database, ctrl.Result, error) {

         // EnsureMariaDBAccount should probably be here now also

	db := mariadbv1.NewDatabaseForAccount(
		instance.Spec.DatabaseInstance, // mariadb/galera service to target
		instance.Name,                  // name used in CREATE DATABASE in mariadb
		instance.Name,                  // CR name for MariaDBDatabase
		instance.Spec.DatabaseAccount,  // CR name for MariaDBAccount
		instance.Namespace,             // namespace
	)

	// create or patch the DB
	ctrlResult, err := db.CreateOrPatchAll(ctx, h)

	if err != nil {
		instance.Status.Conditions.Set(
                // ... condition stuff
	if (ctrlResult != ctrl.Result{}) {
                // ... condition stuff

	ctrlResult, err = db.WaitForDBCreated(ctx, h)

	if err != nil {
		instance.Status.Conditions.Set(
                // ... more condition stuff
	if (ctrlResult != ctrl.Result{}) {
                // ... more condition stuff


	// update Status.DatabaseHostname, used to config the service
	instance.Status.DatabaseHostname = db.GetDatabaseHostname()
	instance.Status.Conditions.MarkTrue(condition.DBReadyCondition, condition.DBReadyMessage)
	return db, ctrlResult, nil
}

```

everything is all together.  Why dont we just mash it up together like this?

```
func (r *KeystoneAPIReconciler) ensureDB(
	ctx context.Context,
	h *helper.Helper,
	instance *keystonev1.KeystoneAPI,
) (*mariadbv1.Database, ctrl.Result, error) {

	_, _, err := mariadbv1.EnsureMariaDBAccount(
		ctx, helper, instance.Spec.DatabaseAccount,
		instance.Namespace, false, "keystone",
	)
	if err != nil {
             / /condition stuff

	db, ctrlResult, err := mariadbv1.SetupNewDatabase(
		instance.Spec.DatabaseInstance, // mariadb/galera service to target
		instance.Name,                           // name used in CREATE DATABASE in mariadb
		instance.Name,                        // CR name for MariaDBDatabase
		instance.Spec.DatabaseAccount,  // CR name for MariaDBAccount
		instance.Namespace,             // namespace
        )

	if err != nil {
		instance.Status.Conditions.Set(
                // ... condition stuff
	if (ctrlResult != ctrl.Result{}) {
                // ... condition stuff

	// update Status.DatabaseHostname, used to config the service
	instance.Status.DatabaseHostname = db.GetDatabaseHostname()
	instance.Status.Conditions.MarkTrue(condition.DBReadyCondition, condition.DBReadyMessage)
	return db, ctrlResult, nil
}

```

I'd feel more confident going with that approach once we go through every downstream controller and we can confirm all of them do these three steps together in all cases now.

api/v1beta1/mariadbdatabase_funcs.go
zzzeek: good call. i guess I had at some point added the MariadbAccount stuff in the middle

api/test/helpers/harnesses.go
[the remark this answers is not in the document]
gibizer: OK

api/v1beta1/mariadbdatabase_funcs.go
gibizer: OK

api/v1beta1/mariadbdatabase_funcs.go
gibizer: > I'd feel more confident going with that approach once we go through every downstream controller and we can confirm all of them do these three steps together in all cases now,

totally agree, lets try to do this as a separate refactor.

PR #189, opened 2024-02-07
dont use logForObject with nil pointer

zzzeek:
the d.account field is known to be nil here, so logforobject will crash on that.  use a plain info log


gibizer (approved):
Thanks

stuggi (approved):
/lgtm

PR #204, opened 2024-03-01
Ensure golangci-lint runs on all files

gibizer:
This needed the following adjustments:
* ignore revive dot-imports rule on ginkgo and gomega as dot import
  there is the recommended practice
* fix a bug that is shown when error flow is indented that when removing
  the finalizer fails that failure is ignored.
* various small style fixes found by the linter


abays (approved):
/lgtm

PR #205, opened 2024-03-04
enhance RunURLAssertSuite to support non-DB managing controllers

zzzeek:
To support isolated testing controllers like ironicapi_controller.go, which renders MySQL URLs from a MariaDBAccount but does not itself manage the creation of the MariaDBAccount, add new RunURLOnlyAssertSuite which does not wait for finalizers to set up; instead, the assertURL function should use Eventually() to wait for the URL change to take place.  This is so that the check
for URL being updated can itself wait, without having to rely on waiting for finalizers to be set up, as these controllers don't actually manage finalizers.

### DIFF
--- a/api/test/helpers/harnesses.go
+++ b/api/test/helpers/harnesses.go
@@ -250,8 +250,24 @@ func (harness *MariaDBTestHarness) RunBasicSuite() {
 			}
 
 			harness.UpdateAccount(newAccountName)
+			harness.mariaDBHelper.SimulateMariaDBAccountCompleted(newAccountName)
 
-			harness.runAccountUpdateWithWait(oldAccountName, newAccountName)
+			mariaDBHelper.Logger.Info(
+				fmt.Sprintf("Service should move to run fully off MariaDBAccount %s and remove finalizer from %s",
+					newAccountName, oldAccountName),
+			)
+
+			// finalizer is attached to new account
+			Eventually(func() []string {
+				newMariadbAccount := mariaDBHelper.GetMariaDBAccount(newAccountName)
+				return newMariadbAccount.Finalizers
+			}, timeout, interval).Should(ContainElement(harness.finalizerName))
+
+			// finalizer removed from old account
+			Eventually(func() []string {
+				oldMariadbAccount := mariaDBHelper.GetMariaDBAccount(oldAccountName)
+				return oldMariadbAccount.Finalizers
+			}, timeout, interval).ShouldNot(ContainElement(harness.finalizerName))
 
 			// CreateOrPatchDBByName will add a label referring to the database
 			Eventually(func() string {
@@ -322,7 +338,8 @@ func (harness *MariaDBTestHarness) RunBasicSuite() {
 }
 
 // RunURLAssertSuite asserts that a database URL is set up with the correct
-// username and password, and that this is updated when the account changes
+// username and password, and that this is updated when the account changes.
+// account change is detected via finalizer
 func (harness *MariaDBTestHarness) RunURLAssertSuite(assertURL assertsURL) {
 	When(fmt.Sprintf("The %s service is fully running", harness.description), func() {
 		BeforeEach(func() {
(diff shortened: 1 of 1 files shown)


zzzeek:
@gibizer this is just an improvement to the harness function that asserts the URL is set or is updated.  I needed this for controllers like ironicapi_controller / ironicconductor_controller.   right now I can just hand-write the tests in those controllers but if this can just go in very quick , that would help.  right now only placement is the only thing merged that's using this

api/test/helpers/harnesses.go
```diff
@@ -321,34 +337,48 @@ func (harness *MariaDBTestHarness) RunBasicSuite() {
...
 
 			oldAccountName := types.NamespacedName{
 				Name:      "some-old-account",
 				Namespace: harness.namespace,
 			}
 
+			k8sClient := mariaDBHelper.K8sClient
+
+			// create MariaDBAccount / secret ahead of time, to suit controllers
+			// that dont directly do EnsureMariaDBAccount
```
gibizer, line 371: I assume that eventually all operators using MariadbAccount will also use EnsureMariaDBAccount. Am I correct? Or EnsureMariaDBAccount is just a temporary step that we will remove from all operators eventually?

api/test/helpers/harnesses.go
```diff
@@ -321,34 +337,48 @@ func (harness *MariaDBTestHarness) RunBasicSuite() {
...
+}
+
+// RunURLAssertSuite asserts that a database URL is set up with the correct
+// username and password, and that this is updated when the account changes.
+// account change is detected via finalizer
 func (harness *MariaDBTestHarness) RunURLAssertSuite(assertURL assertsURL) {
+	harness.runURLAssertSuite(assertURL, true)
+}
+
+func (harness *MariaDBTestHarness) runURLAssertSuite(assertURL assertsURL, useFinalizers bool) {
```
gibizer, line 354: I'm I correct that eventually all operators will add finalizers for the account they use? If so this bool will be eventually removed and the test logic can be simplified back to the original.

gibizer (commented):
I'm not 100% sure this edge case handling is just temporal. If it is just temporal then I'm OK to merge it and revert it later. However if this is expected to be permanent difference between operators then I need explanation why we diverge.

api/test/helpers/harnesses.go
zzzeek: so what is going on is that some test suites are testing the primary controller that does the NewDatabaseForAccount call, which will add finalizers to MariaDBAccount and will also remove from the "old" account, and this harness prior to this commit is relying entirely on these finalizers to determine when it's safe to test that the URL has been generated.    

However, we have a need to test MariaDBAccount related features for other controllers, that do not do NewDatabaseForAccount directly, but they still render MySQL database URLs.  They are just consumers of a MariaDBAccount that would have been created for them and they do not add finalizers.    Examples include ironicapi_controller and ironicconductor_controller, other operators like Heat and Designate have controllers like this as well that also might be doing similar things.  As you know, the functional test suites for each of these controllers only set up the single thing they are testing, like an IronicAPI object and not the parent Ironic instance that would do that work, so the suite fails because the finalizers arent set up and the harness does not have its indication that it's safe to see if the URL was updated.

So in order to get this harness to work as is for those controllers, I was able to manually add finalizer logic into the tests themselves, but then you have a whole lot of extraneous things going on in the consuming controller test, and the point of this suite is to have it be as minimal as possible in those suites.

So what works here is to remove the checks for the finalizers and strictly wait on the URL itself being updated.  so in the ironic API/conductor tests I would be able to do this :

```go
	mariadbSuite.RunURLOnlyAssertSuite(func(accountName types.NamespacedName, username string, password string) {
			Eventually(func(g Gomega) {
				configDataMap := th.GetSecret(ironicNames.APIConfigSecretName)

				conf := configDataMap.Data["ironic.conf"]

				g.Expect(string(conf)).Should(
					ContainSubstring(fmt.Sprintf("connection=mysql+pymysql://%s:%s@%s/ironic?read_default_file=/etc/my.cnf",
						username, password, instance.Spec.DatabaseHostname)))
			}).Should(Succeed())
       })
```

that is, do the wait right on getting the config map and looking at the URL directly, there is no assertion or wait for finalizers.

So - to the point I think you are making in the first place, "why two methods", I agree with you, the best solution would be that the finalizer logic is just removed here and that all consuming test suites need to use Eventually.

We can go with that instead.  It just means we'd have to immediately merge a fix to placement in order for it to work here with the newer mariadb release.

that's probably OK because we'd be merging the newer config hash to get there anyway, right?

zzzeek: not to mention it's possibly not even totally safe to rely 100% on just the finalizer state, since that's not the same as the configmap update logic.      

so can I just change the one method and placement will be fine as it would need a go.mod update anyway

zzzeek: ive proposed the change in placement at https://github.com/openstack-k8s-operators/placement-operator/pull/171

api/test/helpers/harnesses.go
zzzeek: we are likely going to always have EnsureMariaDBAccount however it might change to do just a check and not an actual create.      then also we might consolidate the EnsureMariaDBAccont / NewDatabaseForAccount / WaitForCreated into one API call but we're not there yet.

zzzeek: but yes every operator uses EnsureMariaDBAccount

api/test/helpers/harnesses.go
gibizer: thanks for the explanation. It  make sense

api/test/helpers/harnesses.go
gibizer: OK

gibizer (approved):

PR #206, opened 2024-03-08
report err on failure to remove finalizer from MariaDBDatabase

zzzeek:


controllers/mariadbaccount_controller.go
```diff
@@ -370,12 +370,17 @@ func (r *MariaDBAccountReconciler) reconcileDelete(
 		// first, remove finalizer from the MariaDBDatabase instance
 		if controllerutil.RemoveFinalizer(mariadbDatabase, fmt.Sprintf("%s-%s", helper.GetFinalizer(), instance.Name)) {
 			err = r.Update(ctx, mariadbDatabase)
+
+			if err != nil {
```
gibizer, line 374: This can be refined.  If the error is NotFound then we can continue removing the self finalizer.

controllers/mariadbaccount_controller.go
```diff
@@ -434,6 +439,11 @@ func (r *MariaDBAccountReconciler) reconcileDelete(
 			// remove finalizer from the MariaDBDatabase instance
 			if controllerutil.RemoveFinalizer(mariadbDatabase, fmt.Sprintf("%s-%s", helper.GetFinalizer(), instance.Name)) {
 				err = r.Update(ctx, mariadbDatabase)
+
+				if err != nil {
```
gibizer, line 443: ditto, we can ignore NotFound

controllers/mariadbaccount_controller.go
zzzeek: OK do that for both ?

zzzeek: like this right?

```
				if err != nil && !k8s_errors.IsNotFound(err) {
					return ctrl.Result{}, err
				}

```

PR #210, opened 2024-03-12
apply finalizers to Secret for MariaDBAccount

zzzeek:
when a controller is reconciling the delete of its own CR as well as its MariaDBAccount, the lookup for MariaDBDatabase/MariaDBAccount must
succeed so that db.DeleteFinalizer can be called.    This uses
GetDatabaseByNameAndAccount which is required to locate all three
of MariaDBDatabase, MariaDBAccount, and Secret, else notfound is
returned.

therefore, prevent the Secret from being prematurely removed from the cluster by adding a calling CR finalizer to it, the same as it's added to the MariaDBAccount.

Prior to this change, the GetDatabaseByNameAndAccount can return a not-found even though the MariaDBAccount exists, causing
the calling controller to skip it and delete its own CR, leaving the MariaDBAccount and MariaDBDatabase dangling.

the issue can be reproduced by building up an openstack env with keystone / galera / memcached / rabbitmq, then deleting the namespace. k8s will delete the secrets more quickly than it can remove the CRs since they are reconciling the delete.

### DIFF
--- a/api/v1beta1/mariadbdatabase_funcs.go
+++ b/api/v1beta1/mariadbdatabase_funcs.go
@@ -356,18 +356,18 @@ func (d *Database) CreateOrPatchAll(
 		return ctrl.Result{RequeueAfter: time.Second * 5}, nil
 	}
 
-	opAcc, errAcc := CreateOrPatchAccount(
-		ctx, h, mariaDBAccount,
+	opAcc, err := createOrPatchAccountAndSecret(
+		ctx, h, mariaDBAccount, nil,
 		map[string]string{
 			"mariaDBDatabaseName": d.name,
 		},
 	)
 
-	if errAcc != nil {
+	if err != nil {
 		return ctrl.Result{}, util.WrapErrorForObject(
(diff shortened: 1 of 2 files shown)


api/v1beta1/mariadbdatabase_funcs.go
```diff
@@ -678,21 +687,117 @@ func DeleteUnusedMariaDBAccountFinalizers(
...
+			trueVal := true
+
+			_, ok1 := accountSecret.Data[DatabasePasswordSelector]
+			_, ok2 := accountSecret.StringData[DatabasePasswordSelector]
+			if !ok1 && !ok2 {
+				err := fmt.Errorf("field %s not found in Secret %s", DatabasePasswordSelector, accountSecret.Name)
+				return err
+			}
+
+			accountSecret.Immutable = &trueVal
```
gibizer, line 770: nit: ` accountSecret.Immutable = ptr.To(true)`

gibizer (commented):
Looks good to me. I prefer this over #209

gibizer (approved):

PR #211, opened 2024-03-13
remove now-unused mariadb funcs

zzzeek:
every operator is now updated on the newer functions (octavia is still trying to merge), remove all the old functions and old "legacy" codepaths, as a first pass to making this module a little clearer.


zzzeek:
OK octavia has to merge then to be updated.  ill see if i can poke along the kuttl tests

PR #213, opened 2024-04-02
Re-init conditions each reconcile

fmount:
This patch follows the same pattern applied to the other operators, where we re-init the `conditions` at each reconcile loop. `Conditions` are re-evaluated and updated, keeping the `LastTransitionTime` for those that haven't changed (it avoids the transition from `True` to `Unknown` to `True` again).
In addition, the `observedGeneration` field is introduced, and it is used by the `openstack-operator` to check the `IsReady()` function for a particular `CR` in case a minor update is triggered. All the conditions are evaluated during the main `Reconcile` loop ( or the `reconcileNormal` function in some circumstances), hence the main `ReadyCondition` is updated within the same flow. The defer function still updates the `Resource` and `Mirror` the condition to the top-level `CR`.

Jira: [OSPRH-5919](https://issues.redhat.com//browse/OSPRH-5919)
Jira: [OSPRH-5698](https://issues.redhat.com/browse/OSPRH-5698)


controllers/mariadbaccount_controller.go
```diff
@@ -82,45 +82,44 @@ func (r *MariaDBAccountReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
 
-		instance.Status.Conditions.Init(&cl)
+	instance.Status.Conditions.Init(&cl)
 
-		// Register overall status immediately to have an early feedback e.g. in the cli
-		return ctrl.Result{}, nil
-	}
-
-	if instance.DeletionTimestamp.IsZero() {
+	if instance.DeletionTimestamp.IsZero() || isNewInstance {
```
zzzeek, line 122: what's the scenario where deletion timestamp is non-zero but no conditions were present?  why would we do a reconcilecreate if deletion timestamp is nonzero, doesnt that mean it was marked as deleted?

dprince: I think it would be where it was deleted very quickly. We have this logic in all our operators.

dprince (approved):

PR #216, opened 2024-04-09
use published service hostname for db/account jobs

zzzeek:
also runs a check inside database.sh / account.sh to ensure the command succeeded

### DIFF
--- a/main.go
+++ b/main.go
@@ -143,8 +143,9 @@ func main() {
 	}
 
 	if err = (&controllers.MariaDBAccountReconciler{
-		Client: mgr.GetClient(),
-		Scheme: mgr.GetScheme(),
+		Client:  mgr.GetClient(),
+		Kclient: kclient,
+		Scheme:  mgr.GetScheme(),
 	}).SetupWithManager(mgr); err != nil {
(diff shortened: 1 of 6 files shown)


main.go
```diff
@@ -143,8 +143,9 @@ func main() {
 	}
 
 	if err = (&controllers.MariaDBAccountReconciler{
-		Client: mgr.GetClient(),
-		Scheme: mgr.GetScheme(),
+		Client:  mgr.GetClient(),
+		Kclient: kclient,
```
dciabrin, line 147: Is that needed here? Or is it an additional fix?

zzzeek: it's both!   that kclient is needed for the service lookup that happens inside of here: https://github.com/openstack-k8s-operators/mariadb-operator/blob/00db0f83ca31760c8a6af815082c913d54a556d4/controllers/mariadbaccount_controller.go#L251    it was nil pointer crashing without it

zzzeek: im not sure why the kclient thing wasnt there already since I would have copy-pasted from the mariadbdatabase controller when i first made this, maybe kclient entry was added for mariadb controller after the fact

gibizer (commented):
as far as follow this this look OK

abays (approved):
/lgtm

PR #217, opened 2024-04-11
Update lib-common to contain .generation fix

mrkisaolamb:


gibizer:
> Update lib-common to contain .generation fix

From which module?

gibizer (approved):
We need this as there are test cases defined in mariadb-operator that are called from the service operators and the shared uses a lib-common helper that is updated with .observedGeneration support.

gibizer:
 prow should work now

stuggi (approved):
/lgtm

PR #218, opened 2024-04-19
Full cluster restart when required by a config update

dciabrin:
Configuration updates are normally rolled out one pod after the other, but there are times where this cannot work. For instance, updating galera from non-TLS to TLS require all pods to have access to the CA before they can start mysql and rejoin a cluster, otherwise, TLS communication won't work.

Add logic to track specific configuration changes that require a full cluster restart to be taken into account. When the controller detects such change, it orchestrates the update as follows:
  . update the statefulset to stop all pods
  . when all pods are stopped, update the statefulset another time
    with the updated config and allow all pods to restart

This mechanism can be reused for future features like e.g. updating SST method from rsync to mariabackup, or performing a major upgrade or mariadb.

JIRA: [OSPRH-5294](https://issues.redhat.com//browse/OSPRH-5294)


zzzeek (approved):

stuggi:
@dciabrin could you please rebase this PR

PR #229, opened 2024-06-12
Make Galera service A/P

dciabrin:
The service created for a Galera cluster always balances traffic to all available Galera replicas, which means the cluster is effectively configured as an A/A service.

In order to limit the burden on clients of supporting A/A semantics, change the way the service is configured to act as A/P instead.

A new callback script is introduced in pods. It is called by mysqld automatically every time the Galera library detects a change in the cluster (node join, crash, network partition...). This callback script reconfigures the service CR via label selectors to drive traffic to a single pod at a time.

Jira: [OSPRH-7405](https://issues.redhat.com//browse/OSPRH-7405)

### DIFF
--- a/controllers/galera_controller.go
+++ b/controllers/galera_controller.go
@@ -426,6 +426,11 @@ func (r *GaleraReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
 			Resources: []string{"pods"},
 			Verbs:     []string{"create", "get", "list", "watch", "update", "patch", "delete"},
 		},
+		{
+			APIGroups: []string{""},
+			Resources: []string{"services"},
+			Verbs:     []string{"get", "list", "update", "patch"},
+		},
 	}
 	rbacResult, err := common_rbac.ReconcileRbac(ctx, helper, instance, rbacRules)
 	if err != nil {
@@ -472,7 +477,7 @@ func (r *GaleraReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
 	headless := &corev1.Service{ObjectMeta: pkghl.ObjectMeta}
 	op, err := controllerutil.CreateOrPatch(ctx, r.Client, headless, func() error {
 		headless.Spec = pkghl.Spec
-		err := controllerutil.SetOwnerReference(instance, headless, r.Client.Scheme())
+		err := controllerutil.SetControllerReference(instance, headless, r.Client.Scheme())
 		if err != nil {
 			return err
 		}
@@ -488,8 +493,17 @@ func (r *GaleraReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
 	pkgsvc := mariadb.ServiceForAdoption(instance, "galera", adoption)
 	service := &corev1.Service{ObjectMeta: pkgsvc.ObjectMeta}
 	op, err = controllerutil.CreateOrPatch(ctx, r.Client, service, func() error {
+		// NOTE(dciabrin) We deploy Galera as an A/P service (i.e. no multi-master writes)
+		// by setting labels in the service's label selectors.
+		// This label is dynamically set based on the status of the Galera cluster,
+		// so in this CreateOrPatch block we must reuse whatever is present in
+		// the existing service CR in case we're patching it.
+		activePod, present := service.Spec.Selector[mariadb.ActivePodSelectorKey]
 		service.Spec = pkgsvc.Spec
-		err := controllerutil.SetOwnerReference(instance, service, r.Client.Scheme())
+		if present {
+			service.Spec.Selector[mariadb.ActivePodSelectorKey] = activePod
+		}
+		err := controllerutil.SetControllerReference(instance, service, r.Client.Scheme())
 		if err != nil {
 			return err
 		}
--- a/pkg/mariadb/const.go
+++ b/pkg/mariadb/const.go
@@ -3,4 +3,7 @@ package mariadb
 const (
 	// ServiceName -
 	ServiceName = "mariadb"
+
+	// ActivePodSelectorKey - Selector key used to configure A/P service behavior
+	ActivePodSelectorKey = "statefulset.kubernetes.io/pod-name"
 )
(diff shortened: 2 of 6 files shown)


dciabrin:
@zzzeek my plan is to do that in a subsequent PR, to avoid any CRD change that would force a bump in openstack-operator.

stuggi:
> @zzzeek my plan is to do that in a subsequent PR, to avoid any CRD change that would force a bump in openstack-operator.

Depending on the CRD change, I think we should do it now.

controllers/galera_controller.go
```diff
@@ -488,7 +496,16 @@ func (r *GaleraReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
...
+		// by setting labels in the service's label selectors.
+		// This label is dynamically set based on the Galera cluster's status,
+		// so in this CreateOrPatch block we must reuse whatever is present in
+		// the existing service CR in case we're patching it.
+		activePod, present := service.Spec.Selector[ActivePodSelectorKey]
 		service.Spec = pkgsvc.Spec
+		if present {
+			service.Spec.Selector[ActivePodSelectorKey] = activePod
+		}
 		err := controllerutil.SetOwnerReference(instance, service, r.Client.Scheme())
```
stuggi, line 509: Is there a reason why the instance is not set as the Controller (same for L483)? currently the controller won't reconcile if the service gets deleted.

I guess we need it now for this PR as otherwise the controller will reconcile on the selector update and returns it to what it think it should be?

With this, there is a scenario where we'd loose the a-p selector. If someone deletes the service (for whatever reason), first of all the controller currently won't reconcile and re-create it. So a user would have to restart the mariadb operator, or do some other change to trigger a reconcile, but in this scenario, we don't have the active pod selector as it is a new service and there won't be a change to the galera cluster and the script won't run. With this we are in the multi master config.

Another approach could be (not sure yet which one is better) that the script adds an annotation to the statefulset on which one is the master (I guess on a pod restart it will always run at least once?), the controller would then reconcile and can update the service.
There is also a downside in this approach, the operator needs to run. If the mariadb-operator is down when there is a galera change, those would not be reflected properly. This could happen e.g. if the operator and the deployment pod which was the master are located on a worker which failed. Depending on how log it takes to re-schedule the operator the db would be not available ... so ... maybe it can be a mix, like checking if there is already the statefulset and take the info from there

dciabrin: > Is there a reason why the instance is not set as the Controller (same for L483)? currently the controller won't reconcile if the service gets deleted.

I'm not sure I got this one. Isn't it what lines 509 and 483 supposed to do? When I do e.g. `oc describe service openstack`, I can see a label `owner=mariadb-operator`.

> 
> I guess we need it now for this PR as otherwise the controller will reconcile on the selector update and returns it to what it think it should be?

Hmmm you're correct we'd lose this label in this case.

> 
> With this, there is a scenario where we'd loose the a-p selector. If someone deletes the service (for whatever reason), first of all the controller currently won't reconcile and re-create it. So a user would have to restart the mariadb operator, or do some other change to trigger a reconcile, but in this scenario, we don't have the active pod selector as it is a new service and there won't be a change to the galera cluster and the script won't run. With this we are in the multi master config.

You're correct, the label won't be recreated automatically. Restarting a pod would be sufficient for galera of a cluster change and to recreate the right label, although that would be a manual action.

> 
> Another approach could be (not sure yet which one is better) that the script adds an annotation to the statefulset on which one is the master (I guess on a pod restart it will always run at least once?), the controller would then reconcile and can update the service. There is also a downside in this approach, the operator needs to run. If the mariadb-operator is down when there is a galera change, those would not be reflected properly. This could happen e.g. if the operator and the deployment pod which was the master are located on a worker which failed. Depending on how log it takes to re-schedule the operator the db would be not available ... so ... maybe it can be a mix, like checking if there is already the statefulset and take the info from there

So I'm against trying to duplicate the state in the CR's status because I don't want to ensure that the states are in sync. However we could handle this deletion case by selecting any one of the healthy pod as the current active endpoint. This wouldn't conflict with the script as it only performs conditional writes.

dciabrin: > > > Is there a reason why the instance is not set as the Controller (same for L483)? currently the controller won't reconcile if the service gets deleted.
> > 
> > 
> > I'm not sure I got this one. Isn't it what lines 509 and 483 supposed to do? When I do e.g. `oc describe service openstack`, I can see a label `owner=mariadb-operator`.
> 
> They are adding an owner, but not as the controller. `Owns(&corev1.Service{}).` in the `NewControllerManagedBy()` will only reconcile on the ones are also the controller, like in https://github.com/openstack-k8s-operators/mariadb-operator/blob/main/controllers/galera_controller.go#L447 .
> 

Oh I see now... Thanks for pointing that out. I don't see why we shouldn't own it, I am going to update
the PR.

> No, I am not saying to store it in the CR status. Right now the script updates the service/endpoint. it could also add an annotation to the statefulset object with the information. And the controller here takes it from there instead of the service. since the statefulset remains in the mentioned scenario, the information would survive.
> 
> But we could handle this in a follow up. just wanted to bring up that there are situations where we may get back to the a-a config.

I think it's sufficient to always set the first pod as the current active endpoint, as 1) if that information becomes incorrect, it is guaranteed to get updated by galera as soon as it detects that this galera node is unresponsive. 2) if it is a healthy pod, it can be used as the active endpoint anyway, and the script will still be able to perform failover next time it is needed.

controllers/galera_controller.go
```diff
@@ -800,6 +814,7 @@ func (r *GaleraReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&corev1.Endpoints{}).
 		Owns(&corev1.ConfigMap{}).
 		Owns(&corev1.ServiceAccount{}).
+		Owns(&corev1.Service{}).
```
stuggi, line 817: already in L813

dciabrin:
As discussed with @stuggi, we now do SetControllerReference on all Service CRs to get notified of deleting and react appropriately.

stuggi (approved):
/lgtm

PR #231, opened 2024-06-18
update finalzier to CRD doc

stuggi:
sync finalizers across the operators

Depends-on: https://github.com/openstack-k8s-operators/lib-common/pull/519


api/v1beta1/mariadbdatabase_funcs.go
```diff
@@ -458,13 +458,12 @@ func (d *Database) DeleteFinalizer(
 	h *helper.Helper,
 ) error {
 
-	secretFinalizer := fmt.Sprintf("mariadb.openstack.org/%s", h.GetFinalizer())
```
zzzeek, line 461: Just to confirm, what I'm seeing in https://github.com/openstack-k8s-operators/lib-common/pull/519/files combined with this change indicates that the name of the finalizer will change from:

`mariadb.openstack.org/<kind>`

to 

`openstack.org/<kind>`

and I assume that's what we want to have for all operators.

zzzeek (approved):
LGTM

api/v1beta1/mariadbdatabase_funcs.go
stuggi: yes, correct, want to align them to `openstack.org/<kind>` . we also need to consider the max length of 63char. so omit a prefix per operator can be omitted I think

PR #247, opened 2024-07-17
Align Makefile and fix golangci errors

stuggi:
so this is just a PR to align the Makefile and how functional tests run in CI with the other operators and fix things to make golangci happy.

Depends-On: https://github.com/openshift/release/pull/54445


api/v1beta1/galera_webhook.go
```diff
@@ -121,6 +123,11 @@ func (spec *GaleraSpecCore) ValidateCreate(basePath *field.Path) (admission.Warn
 func (r *Galera) ValidateUpdate(old runtime.Object) (admission.Warnings, error) {
 	galeralog.Info("validate update", "name", r.Name)
 
+	oldGalera, ok := old.(*Galera)
```
stuggi, line 126: to not set  `_` for old I added this.

api/test/helpers/harnesses.go
```diff
@@ -62,7 +62,7 @@ type MariaDBTestHarness struct {
 	mariaDBHelper      *TestHelper
 	timeout            time.Duration
 	interval           time.Duration
-	runUpdate          bool
+	//runUpdate          bool - unused
```
abays, line 65: Should we just remove this line entirely if it's not used?

stuggi: I left it as I wanted to check with @zzzeek

zzzeek: I would assume the additonal linter steps added here found that this was already unused, is that correct?   if it's not used then it's fine

zzzeek (approved):

api/test/helpers/harnesses.go
stuggi: yes it was reported to not be used. I just left it if you are planning for additional functionality which will need it. I'll push an update to remove it

zzzeek: if i wanted it back I'd add it again :)   no idea what it was for

api/test/helpers/harnesses.go
```diff
@@ -62,7 +62,6 @@ type MariaDBTestHarness struct {
 	mariaDBHelper      *TestHelper
 	timeout            time.Duration
 	interval           time.Duration
-	runUpdate          bool
```
zzzeek, line 65: cc @mrkisaolamb this was committed in 0c306d23 was this a merge artifact of some kind?

zzzeek (approved):
if you did a grep on everything else then this should be fine, if some other controller has this flag in their test suite and then won't build, it seems to do nothing, so we'd just bump that controiller to not use it

api/test/helpers/harnesses.go
mrkisaolamb: yes this is merge artifact

mrkisaolamb (approved):

PR #254, opened 2024-07-31
Fix non-Primary checks in liveness probe

dciabrin:
The liveness probe did not parse non-Primary condition out of the mysql CLI command. Consequently, the liveness did not fail whe a galera node was disconnected from the primary partition and the galera pod could not restart automatically, leading to long delays before restart or sometimes full cluster disruption.

Fix the way probes are handled and refactors bits to allow more precise conditions in startup/readiness/liveness probes.

Jira: [OSPRH-8862](https://issues.redhat.com//browse/OSPRH-8862)

### DIFF
--- a/templates/galera/bin/mysql_probe.sh
+++ b/templates/galera/bin/mysql_probe.sh
@@ -1,27 +1,37 @@
 #!/bin/bash
-set -eu
+set -u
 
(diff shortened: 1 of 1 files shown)


stuggi (approved):
/lgtm

PR #255, opened 2024-08-05
Do not start galera as joiner with 1-replica cluster

dciabrin:
The mariadb operator checks for available pods in the galera statefulset to determine whether to start mysqld as a bootstrap or a joiner node on all the pods that remain to be started.

When galera is deployed as a 1-replica cluster (e.g. in CI), there is a small time window after the statefulset has been probed and galera marked as 'bootstrapped', where the single pod can crash before being probed. If so, the operator will try to restart the pod as a 'joiner', which is invalid.

Add a specific check for 1-replica deployments, so that the operator bails out and requeue the event when a pod is identified as a joiner. This allows the operator to reprobe the galera state restart the pod correctly, in order to avoid an unecessary error in the logs.

Jira: [OSPRH-7821](https://issues.redhat.com//browse/OSPRH-7821)


stuggi (approved):
/lgtm

PR #256, opened 2024-08-09
[OSPRH-9371] Fix VerifySecret so it requeues when necessary

abays:
We've come to realize three things recently:

1. Returning an error to the reconciler along with a `RequeueAfter` will cause the reconciler to abandon the explicit requeue and only consider the error [1].
2. The `VerifySecret` function in `lib-common` was returning an error along with a requeue, thus contributing to the problem listed in item 1. [2]
3. Also, the error `VerifySecret` was returning was _not_ a "not found" error, but a generic k8s error -- so it wouldn't be properly handled by callers looking for a "not found" error. [3]

Now `VerifySecret` only returns a non-empty `ctrl.Result{}` if the `Secret` is missing. [4]

Jira: https://issues.redhat.com/browse/OSPRH-9371

[1] https://github.com/openstack-k8s-operators/keystone-operator/pull/459
[2] https://github.com/openstack-k8s-operators/lib-common/blob/ad3edb6e67ab738cfdbf4e2e78ac403784a08ab7/modules/common/secret/secret.go#L423-L425
[3] https://github.com/openstack-k8s-operators/lib-common/blob/ad3edb6e67ab738cfdbf4e2e78ac403784a08ab7/modules/common/secret/secret.go#L425
[4] https://github.com/openstack-k8s-operators/lib-common/blob/579da98fa7a62b70b43cdba92e82d6497afe8835/modules/common/secret/secret.go#L425-L427


stuggi (approved):
/lgtm

PR #260, opened 2024-08-28
add multi-master option to galera

zzzeek:
This change adds the new value enableMultiMaster to the Galera CR defaulting to false; when set to true, the "statefulset.kubernetes.io/pod-name" key is removed from the mariadb service spec, and when set to false it is restored.  This has the effect of publishing the service either under one specific pod for active/passive use, or using any pod for any request which is multi-master use.  the value of the attribute is also propigated out to the status within the MariaDBDatabase CR, which is then consumable by openstack
services in a similar manner as the TLS settings.   Openstack
services will then want to set wsrep_sync_wait accordingly
for AP or MM mode, which will be proposed in followup changes.

Fixes: [OSPRH-7406](https://issues.redhat.com//browse/OSPRH-7406)

### DIFF
--- a/controllers/galera_controller.go
+++ b/controllers/galera_controller.go
@@ -509,16 +509,23 @@ func (r *GaleraReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
 		// service get stuck.
 		controllerutil.AddFinalizer(service, helper.GetFinalizer())
 
-		// NOTE(dciabrin) We deploy Galera as an A/P service (i.e. no multi-master writes)
-		// by setting labels in the service's label selectors.
-		// This label is dynamically set based on the status of the Galera cluster,
-		// so in this CreateOrPatch block we must reuse whatever is present in
-		// the existing service CR in case we're patching it.
-		activePod, present := service.Spec.Selector[mariadb.ActivePodSelectorKey]
-		service.Spec = pkgsvc.Spec
-		if present {
-			service.Spec.Selector[mariadb.ActivePodSelectorKey] = activePod
+		if !instance.Spec.EnableMultiMaster {
+			// NOTE(dciabrin) We deploy Galera as an A/P service (i.e. no multi-master writes)
+			// by setting labels in the service's label selectors.
+			// This label is dynamically set based on the status of the Galera cluster,
+			// so in this CreateOrPatch block we must reuse whatever is present in
+			// the existing service CR in case we're patching it.
+			activePod, present := service.Spec.Selector[mariadb.ActivePodSelectorKey]
+			service.Spec = pkgsvc.Spec
+
+			if present {
+				service.Spec.Selector[mariadb.ActivePodSelectorKey] = activePod
+			}
+		} else {
+			service.Spec = pkgsvc.Spec
+			delete(service.Spec.Selector, mariadb.ActivePodSelectorKey)
 		}
+
 		err := controllerutil.SetControllerReference(instance, service, r.Client.Scheme())
 		if err != nil {
 			return err
--- a/controllers/mariadbdatabase_controller.go
+++ b/controllers/mariadbdatabase_controller.go
@@ -27,6 +27,8 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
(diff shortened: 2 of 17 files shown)


controllers/galera_controller.go
```diff
@@ -509,16 +509,23 @@ func (r *GaleraReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
...
+			// the existing service CR in case we're patching it.
+			activePod, present := service.Spec.Selector[mariadb.ActivePodSelectorKey]
+			service.Spec = pkgsvc.Spec
+
+			if present {
+				service.Spec.Selector[mariadb.ActivePodSelectorKey] = activePod
+			}
+		} else {
+			service.Spec = pkgsvc.Spec
+			delete(service.Spec.Selector, mariadb.ActivePodSelectorKey)
```
zzzeek, line 526: OK so not sure what to do here if someone wanted to flip a cluster into multimaster mode when mariadbdatabases are already present:

1. the status in each mariadbdatabase for "multimaster" would be stale unless they were all notified
2. even if we went and refreshed all mariadbdatabases, thedownstream services consuming the mariadbdatabase would *also* need to be refreshed because they'd have to change wsrep_sync_notify

do we have this problem elsewhere?  what do we do if TLS settings on Galera are changed on a running cluster?

cc @dciabrin @stuggi

stuggi: in general the services watch their mariadbdatabase e.g. https://github.com/openstack-k8s-operators/keystone-operator/blob/main/controllers/keystoneapi_controller.go#L331 and will reconcile when it changes. I think when we change the mariadbdatabase controller to watch the galera instance CR it maps to, it will reconcile and get the multimaster update. With this also the downstream services would reconcile and get the update.

for tls with the DB its bit different. Galera always gets configured to support tls. basically right now you can not disable tls for the db.

zzzeek: ah right, I guess that works

so here I need to get the MariaDBDatabases to react when someone changes this setting, but if they are already watching their Galera then it would just work

stuggi: I think what we have to do to add here https://github.com/openstack-k8s-operators/mariadb-operator/blob/main/controllers/mariadbdatabase_controller.go#L247 to watch the galera cr. then the mariadbdatabase should update its enablemultimaster when the config changes.

zzzeek: yah that's what i was thinking, what do we need to look out for if we do that?

i will try that here and also extend the kuttl test to make sure that happens

stuggi: need to add a watches for the galera which is set on the mysqldatbase as the dbName label https://github.com/openstack-k8s-operators/mariadb-operator/blob/main/controllers/mariadbdatabase_controller.go#L257 . could do similar to what we have e.g. here for the memcached https://github.com/openstack-k8s-operators/keystone-operator/blob/main/controllers/keystoneapi_controller.go#L342-L343 with the memcachedFn. could limit it to the `builder.WithPredicates(predicate.ResourceVersionChangedPredicate{}),` .

zzzeek: it didnt seem to work unless I used a full  `handler.EnqueueRequestsFromMapFunc` approach, which works

controllers/mariadbdatabase_controller.go
```diff
@@ -145,7 +147,7 @@ func (r *MariaDBDatabaseReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 
 	// here we know that Galera exists so add a finalizer to ourselves and to the db CR. Before this point there is no reason to have a finalizer on ourselves as nothing to cleanup.
 	if instance.DeletionTimestamp.IsZero() || isNewInstance { // this condition can be removed if you wish as it is always true at this point otherwise we would returned earlier.
-		if controllerutil.AddFinalizer(dbGalera, fmt.Sprintf("%s-%s", helper.GetFinalizer(), instance.Name)) {
+		if dbGalera.DeletionTimestamp.IsZero() && controllerutil.AddFinalizer(dbGalera, fmt.Sprintf("%s-%s", helper.GetFinalizer(), instance.Name)) {
```
zzzeek, line 150: @dciabrin can you review this line, without this the kuttl tests would run into a stack trace as trying to reconcile the mariadbdatabase after galera were deleted would raise here (this happens more frequently due to the Watch I added in this PR).   note that by checking this, we do continue on to add a finalizer to the mariadbdatabase itself if one was not there already, but I think that would have been there anyway

stuggi:
looks good to me, @dciabrin when you are also happy with it, we could get it in.

PR #264, opened 2024-09-13
[18.0.0-proposed] Add CR name length validation

stuggi:
The galera controller creates StatefulSet for the db to run. This adds a StatefulSet pod's label
"controller-revision-hash": "<statefulset_name>-<hash>" to the pod. The kubernetes label is restricted under 63 char and the revision hash is an int32, 10 chars + the hyphen. The galera controller also adds a suffix '-galera' to the statefulset name. With this the max name must not exceed 46 chars.

Depends-On: https://github.com/openstack-k8s-operators/lib-common/pull/562

Jira: https://issues.redhat.com/browse/OSPRH-8063

Signed-off-by: Martin Schuppert <mschuppert@redhat.com>
(cherry picked from commit 4f04456810c3142660d55599cad0a5717bd8ee3c)


abays (approved):
/lgtm

PR #265, opened 2024-09-13
Check DbRootPassword exist using VerifySecret()

stuggi:
Verification of the password secret should also check for the expected key (DbRootPassword) to exist. This changes this by using the secret.VerifySecret() func.

### DIFF
--- a/controllers/galera_controller.go
+++ b/controllers/galera_controller.go
@@ -539,18 +539,32 @@ func (r *GaleraReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
 	clusterPropertiesEnv := make(map[string]env.Setter)
 
 	// Check and hash inputs
-	secretName := instance.Spec.Secret
 	// NOTE do not hash the db root password, as its change requires
 	// more orchestration than a simple rolling restart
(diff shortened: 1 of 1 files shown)


controllers/galera_controller.go
```diff
@@ -539,18 +539,32 @@ func (r *GaleraReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
...
 	// Check and hash inputs
-	secretName := instance.Spec.Secret
 	// NOTE do not hash the db root password, as its change requires
 	// more orchestration than a simple rolling restart
-	_, _, err = secret.GetSecret(ctx, helper, secretName, instance.Namespace)
+	_, res, err := secret.VerifySecret(
+		ctx,
+		types.NamespacedName{Namespace: instance.Namespace, Name: instance.Spec.Secret},
+		[]string{
+			"DbRootPassword",
```
stuggi, line 548: we probably should add a crd parameter for the password selector, like we have for OSP service users.

stuggi:
related to https://issues.redhat.com/browse/OSPRH-7992

PR #268, opened 2024-09-17
add DeleteDatabaseAndAccountdFinalizers

zzzeek:
the current approach of removing finalizers is dependent on getting at a MariaDBDatabase and a complete MariaDBAccount with a Secret at the
same time to produce a Database object.  At that point, the finalizers
are removed from the sub-elements.   however if the MariaDBAccount
or MariaDBDatabase is missing then nothing gets removed, this is a
bug.  So add new function that looks up all three types of objects
independently of each other and remove finalizers from each, regardless
of which are found or not.


api/v1beta1/mariadbdatabase_funcs.go
```diff
@@ -532,6 +524,73 @@ func (d *Database) GetDatabaseClientConfig(s *tls.Service) string {
...
+	if err != nil && !k8s_errors.IsNotFound(err) {
+		return err
+	} else if err == nil {
+		if databaseAccount.Spec.Secret != "" {
+			dbSecret, _, err := secret.GetSecret(ctx, h, databaseAccount.Spec.Secret, namespace)
+			if err != nil && !k8s_errors.IsNotFound(err) {
+				return err
+			}
+
+			if controllerutil.RemoveFinalizer(dbSecret, h.GetFinalizer()) {
```
zzzeek, line 551: the secret might not be found.   change to `if err == nil and controllerutil.RemoveFinalizer ...`

api/v1beta1/mariadbdatabase_funcs.go
```diff
@@ -532,6 +524,73 @@ func (d *Database) GetDatabaseClientConfig(s *tls.Service) string {
...
+		ctx, h, name, accountName, namespace,
+	)
+	if err != nil && !k8s_errors.IsNotFound(err) {
+		return err
+	}
+
+	mariaDBDatabase, err := GetDatabase(ctx, h, name, namespace)
+	if err != nil && !k8s_errors.IsNotFound(err) {
+		return err
+	} else if err == nil {
```
zzzeek, line 581: remove nesting, change to `else if err == nil and  controllerutil.RemoveFinalizer...`

api/v1beta1/mariadbdatabase_funcs.go
```diff
@@ -532,6 +524,71 @@ func (d *Database) GetDatabaseClientConfig(s *tls.Service) string {
...
+			if err != nil && !k8s_errors.IsNotFound(err) {
+				return err
+			}
+
+			if err == nil && controllerutil.RemoveFinalizer(dbSecret, h.GetFinalizer()) {
+				err := h.GetClient().Update(ctx, dbSecret)
+				if err != nil && !k8s_errors.IsNotFound(err) {
+					return err
+				}
+				util.LogForObject(h, fmt.Sprintf("Removed finalizer %s from Secret %s", h.GetFinalizer(), dbSecret.Name), dbSecret)
```
dciabrin, line 556: Would we log that line if a previous call to DeleteDatabaseAndAccountFinalizers already removed the finalizer for dbSecret, fail a bit later, and force the Openstack client to call DeleteDatabaseAndAccountFinalizers another time?

api/v1beta1/mariadbdatabase_funcs.go
```diff
@@ -532,6 +524,71 @@ func (d *Database) GetDatabaseClientConfig(s *tls.Service) string {
...
+
+	mariaDBDatabase, err := GetDatabase(ctx, h, name, namespace)
+	if err != nil && !k8s_errors.IsNotFound(err) {
+		return err
+	} else if err == nil && controllerutil.RemoveFinalizer(mariaDBDatabase, h.GetFinalizer()) {
+		err := h.GetClient().Update(ctx, mariaDBDatabase)
+		if err != nil && !k8s_errors.IsNotFound(err) {
+			return err
+		}
+		util.LogForObject(h, fmt.Sprintf("Removed finalizer %s from MariaDBDatabase %s", h.GetFinalizer(), mariaDBDatabase.Spec.Name), mariaDBDatabase)
```
dciabrin, line 586: Side question.
Would there be any value in refactoring this pattern in a generic function, and call it once for each object to remove finalizers from? The reason I asking is that this would probably simplify reading this function, and also I can see this pattern being used in other operators, so eventually such a generic function would probably benefit being in libcommon?

api/v1beta1/mariadbdatabase_funcs.go
zzzeek: controllerutil.removeFinalizer would return False if there were no finalizer to remove right?  then the log line would not be emitted

IIUC removeFinalizer modifies the object structure, GetClient().Update() pushes and commits the change in the object DB

dciabrin: Alright thanks I wasn't sure...
https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#RemoveFinalizer for reference.

api/v1beta1/mariadbdatabase_funcs.go
```diff
@@ -532,6 +524,71 @@ func (d *Database) GetDatabaseClientConfig(s *tls.Service) string {
 	return strings.Join(conn, "\n")
 }
 
+// DeleteDatabaseAndAccountFinalizers performs the same tasks as
+// GetDatabaseByNameAndAccount and then database.DeleteFinalizer, but does
+// so such that all individual objects that exist are guaranteed to be updated,
+// even if other objects in the Database combination don't exist.  This
+// includes the MariaDBDatabase, all MariaDBAccount objects and their
+// associated Secret objects.
+func DeleteDatabaseAndAccountFinalizers(
```
gibizer, line 533: it looks good to me. but I will would like to do one more test replacing the copy of this code in https://github.com/openstack-k8s-operators/nova-operator/pull/846/files with actually calling this code and then run the cell deletion tests to see if we still clean up. I hope I can do that test today.

gibizer: I did the test this works for the in flight cell delete use case as well. Thanks.

gibizer (approved):

api/v1beta1/mariadbdatabase_funcs.go
stuggi: not sure if that was the question, `util.LogForObject`, was initially added before you could just get a logger from ctx. The longer term goal was to remove LogForObject and just get a logger as described in https://github.com/openstack-k8s-operators/dev-docs/blob/main/developer.md#logging . a couple of operators have jiras for "T159: Follow best practices for secure error and exception handling", I think we'll review/discuss how logging is done across the operators as part of this.

gibizer:
@dciabrin could  you please check this again? I think we only need an approve from you.

dciabrin:
/approve
Thanks @zzzeek

PR #270, opened 2024-09-27
Do not log an error during resource deletion

dciabrin:
A notify script is being called by the galera library each time it transitions into another state (e.g. Connected -> Synced). The notify script uses a token generated by k8s to communicate with the API server and update the current endpoint for the galera service.

When a galera resource is deleted, the token becomes invalid, so when the notify script is called, it cannot connect to the API server anymore. This is not needed as the resource is being deleted anyway, so handle that case to avoid logging unecessary errors.


abays (approved):
/lgtm

PR #271, opened 2024-09-27
Fix clean up of old attributes when containers are not restarting

dciabrin:
When a pod crashes while in the process of starting, the operator cleans up outdated attributes in the galera CR status.

The operator wrongly assumes that it can probe a container's state as soon as it gets a pod object from the API server, which is not always true (e.g when the pod is in "Pending" state).

Fix the attribute clean up by always checking the state of the pod's container before inspecting its container ID.

Jira: [OSPRH-9411](https://issues.redhat.com//browse/OSPRH-9411)


abays (approved):
/lgtm

PR #274, opened 2024-10-02
Clean up adoption configuration for galera CR

dciabrin:
The Galera CR currently has attributes there were exposed for easing the adoption process (transition of an existing galera cluster to a k8s-managed galera cluster). However this has never been used as it is unneeded with the current adoption implementation.

Remove those flags and simplify the operator accordingly.

Jira: [OSPRH-8058](https://issues.redhat.com//browse/OSPRH-8058)


abays (approved):
/lgtm

PR #276, opened 2024-10-14
Allow intermediate replicas to ease operation

dciabrin:
Currently the mariadb operator has webhooks to enforce the galera replica count to be either 1 or 3, which are the two supported deployments on the OpenstackControlPlane.

However there is a need for more flexibility for day two operations, where it can be useful to allow scaling down replicas temporarily to e.g. 2 (for replacement procedure) or even 0 (to temporarily disable database management).

Replace the kubebuilder annotations with minimum and maximum, and let the galera webhook warn about setting replicas to 2 (as it is not advisable to keep that setting for quorum concerns).


abays (approved):
/lgtm

PR #289, opened 2024-11-20
Improve the failover of galera service

dciabrin:
When a galera node is in the process of shutting down (e.g. during a rolling restart caused by a minor update), the node is unable to serve SQL queries, however it is still connected to clients. This confuses clients who get unexpected SQL status [1] and prevent them from retrying their queries, causing unexpected errors down the road.

Improve the pod stop pre-hook to failover the active endpoint to another pod prior to shutting down the galera server, and kill connected clients to force them to reconnect to the new active endpoint. At this stage, the galera server can be safely shutdown as no client will see its WSREP state update.

Also update the failover script: 1) when no endpoint is available, ensure no traffic is going through any pod. 2) do not trigger a endpoint failover as long as the current endpoint targets a galera node that is still part of the primary partition (i.e. it is still able to serve traffic).

[1] 'WSREP has not yet prepared node for application use'

Jira: [OSPRH-11488](https://issues.redhat.com//browse/OSPRH-11488)

### DIFF
--- a/templates/galera/bin/mysql_shutdown.sh
+++ b/templates/galera/bin/mysql_shutdown.sh
@@ -44,5 +44,16 @@ if curl -s --cacert ${CACERT} --header "Content-Type:application/json" --header 
     done
 fi
 
-log "Shutting down local galera node"
+log "Initiating orchestrated shutdown of the local galera node"
+
+log "Failover service to another available galera node"
+bash $(dirname $0)/mysql_wsrep_notify.sh --status failover
+
+log "Close all active connections to this local galera node"
+# filter out system and localhost connections, only consider clients with a port in the host field
+# from that point, clients will automatically reconnect to another node
+CLIENTS=$(mysql -uroot -p${DB_ROOT_PASSWORD} -nN -e "select id from information_schema.processlist where host like '%:%';")
+echo -n "$CLIENTS" | tr '\n' ',' | xargs mysqladmin -uroot -p${DB_ROOT_PASSWORD} kill
+
+log "Shutdown local server"
 mysqladmin -uroot -p"${DB_ROOT_PASSWORD}" shutdown
--- a/templates/galera/bin/mysql_wsrep_notify.sh
+++ b/templates/galera/bin/mysql_wsrep_notify.sh
@@ -32,18 +32,26 @@ function log_error() {
 function mysql_get_status {
     local name=$1
     mysql -nNE -uroot -p"${DB_ROOT_PASSWORD}" -e "show status like '${name}';" | tail -1
-    if [ $? != 0 ]; then
-        log_error "could not get value of mysql variable '${name}' (rc=$?)"
-        return 1
-    fi
+    local rc=$?
+    [ $rc = 0 ] || log_error "could not get value of mysql variable '${name}' (rc=$rc)"
 }
 
-# Refresh environment variables with the latest WSREP state from mysql
+function mysql_get_members {
+    mysql -nN -uroot -p"${DB_ROOT_PASSWORD}" -e "select node_name from mysql.wsrep_cluster_members;"
+    local rc=$?
+    [ $rc = 0 ] || log_error "could not get cluster members from mysql' (rc=$rc)"
+}
+
+# When optional script parameters are not provided, set up the environment
+# variables with the latest WSREP state retrieved from mysql
 function mysql_probe_state {
-    UUID=$(mysql_get_status wsrep_gcomm_uuid)
-    PARTITION=$(mysql_get_status wsrep_cluster_status)
-    INDEX=$(mysql_get_status wsrep_local_index)
-    SIZE=$(mysql_get_status wsrep_cluster_size)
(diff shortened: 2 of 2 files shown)


stuggi:
looks good to me

gibizer:
> and kill connected clients to force them to reconnect to the new active endpoint. 
Can we make it graceful? Is there a way to handle in progress queries and only kill idle client connections and delaying the pod stop until all the in progress queries are done and therefore all clients are gracefully disconnected?

templates/galera/bin/mysql_shutdown.sh
```diff
@@ -44,5 +44,16 @@ if curl -s --cacert ${CACERT} --header "Content-Type:application/json" --header
...
+log "Initiating orchestrated shutdown of the local galera node"
+
+log "Failover service to another available galera node"
+bash $(dirname $0)/mysql_wsrep_notify.sh --status failover
+
+log "Close all active connections to this local galera node"
+# filter out system and localhost connections, only consider clients with a port in the host field
+# from that point, clients will automatically reconnect to another node
+CLIENTS=$(mysql -uroot -p${DB_ROOT_PASSWORD} -nN -e "select id from information_schema.processlist where host like '%:%';")
+echo -n "$CLIENTS" | tr '\n' ',' | xargs mysqladmin -uroot -p${DB_ROOT_PASSWORD} kill
```
gibizer, line 56: can we make this graceful by only killing the client after it finished the in progress query if any?

dciabrin: sadly no, there's no option in mysql to mark a connection for closing after it finished its processing.

gibizer: I ran couple of rollouts back an forth while polling the keystone API and at least I was not able to hit the case when an ongoing query was interrupted. Actually I was not able to reproduce any API outage during galera rollout after this fix. So  I'm OK to merge this. Thanks @dciabrin for fixing this.

gibizer (approved):

dciabrin:
Rebasing this PR now to fix the kuttl test failure:
``` 
pkg/openstack/galera.go:172:23: cannot use &instance.Spec.NodeSelector (value of type *map[string]string) as map[string]string value in assignment
make: *** [Makefile:147: vet] Error 1
{"component":"entrypoint","error":"wrapped process failed: exit status 2","file":"sigs.k8s.io/prow/pkg/entrypoint/run.go:84","func":"sigs.k8s.io/prow/pkg/entrypoint.Options.internalRun","level":"error","msg":"Error executing test process","severity":"error","time":"2024-11-21T14:06:53Z"} 
```

gibizer (approved):

templates/galera/bin/mysql_wsrep_notify.sh
```diff
@@ -164,6 +175,39 @@ function reconfigure_service_endpoint {
...
+    CURRENT_ENDPOINT=$(echo "$CURRENT_SVC" | parse_output '["spec"]["selector"].get("statefulset.kubernetes.io/pod-name","")')
+    [ $? == 0 ] || return 1
+    if [ "${CURRENT_ENDPOINT}" != "${PODNAME}" ]; then
+        log "Node ${PODNAME} is not the active endpoint. Nothing to be done."
+        return 0
+    fi
+    # select the first available node in the primary partition to be the failover endpoint
+    NEW_ENDPOINT=$(echo "$MEMBERS" | grep -v "${PODNAME}" | head -1)
+    if [ -z "${NEW_ENDPOINT}" ]; then
+        log "No other available node to become the active endpoint."
```
bogdando, line 198: we probably should return with that fact

templates/galera/bin/mysql_wsrep_notify.sh
```diff
@@ -164,6 +175,39 @@ function reconfigure_service_endpoint {
...
+    fi
+
+    NEW_SVC=$(echo "$CURRENT_SVC" | service_endpoint "$NEW_ENDPOINT")
+    [ $? == 0 ] || return 1
+
+    log "Configuring a new active endpoint for service ${SERVICE}: '${CURRENT_ENDPOINT}' -> '${NEW_ENDPOINT}'"
```
bogdando, line 204: ... or this would look like <current> -> (empty)

templates/galera/bin/mysql_wsrep_notify.sh
```diff
@@ -215,6 +271,15 @@ if [ -z "${STATUS}" ]; then
     exit 1
 fi
 
+# Contition: ask for a failover. This should be called when mysql is running
+if echo "${STATUS}" | grep -i -q -e 'failover'; then
+    mysql_probe_state
+    if [ $? != 0 ]; then
+        log_error "Could not probe missing mysql information. Aborting"
```
bogdando, line 278: "... Aborting during failover" would make this msg looking different to when mysql was started

bogdando (commented):
sorry being late to the party

PR #297, opened 2024-12-20
Improve pre-stop hook to avoid spurious logs

dciabrin:
Update the way the pre-stop hook interacts with the mysql server, to avoid logs showing up in k8s events:

 . if no mysql socket is available, do no try to connect.

 . do not stop the server in the pre-hook script, this triggers a
   failed stop in the k8s events.

 . do not call the API server to failover the current endpoint, k8s
   removes the stopping pod automatically, and the remaining galera
   nodes can elect a new endpoint once the pod is stopping.

 . do not call mysqladmin when there are no client connections

Jira: [OSPRH-11392](https://issues.redhat.com//browse/OSPRH-11392)


dciabrin:
/retest-required
registry quay.rdoproject.org timed out

stuggi (approved):
/lgtm

PR #298, opened 2024-12-20
Improve tracking and error reporting of startup probe

dciabrin:
Currently startup probe are scheduled with defaults from k8s (scheduled every 10s, failure threshold of 3). As galera joiner nodes can take a long time to start, this generates unecessary unhealthy events.

Rework how the startup probe work by allowing a single, long probe which internally loops while probe the startup state. Throughout the startup process, keep track of the specific startup phase so in case the startup times out, the probe can log a precise error.

Also rework how joiner nodes are tracked, to fail early in case galera cannot join a primary partition, to avoid the server being stuck until indefinitely until the startup probe times out.

A subsequent commit will provide the ability to override probe settings and timeouts.

Jira: [OSPRH-11392](https://issues.redhat.com//browse/OSPRH-11392)

### DIFF
--- a/pkg/mariadb/const.go
+++ b/pkg/mariadb/const.go
@@ -6,4 +6,7 @@ const (
 
 	// ActivePodSelectorKey - Selector key used to configure A/P service behavior
 	ActivePodSelectorKey = "statefulset.kubernetes.io/pod-name"
+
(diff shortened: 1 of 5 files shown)


dciabrin:
/retest-required
CI job could not build image to run kuttl

PR #304, opened 2025-02-07
Consume Topology CR by reference

fmount:
This patch provides more granular control over Pod placement and scheduling through the Topology CR integration introduced in [1]. In particular it provides:

- a new API parameter (TopologyRef) defined for each Component that allows to reference an existing Topology CRs in the same namespace

- the operator logic that retrieves and processes the referenced Topology CR through the functions provided by lib-common [2]

- enhanced Deployment configuration that incorporates the processed Topology

- kuttl test to verify a referenced topology has been applied to the Deployment and the resulting Pods

Note that webhooks are in place to prevent referencing a Topology from a different namespace (which is not a supported scenario).

Jira: https://issues.redhat.com/browse/OSPRH-13836


tests/kuttl/tests/galera_deploy_topology/01-deploy_galera_topology.yaml
```diff
@@ -0,0 +1,32 @@
...
+---
+apiVersion: topology.openstack.org/v1beta1
+kind: Topology
+metadata:
+  name: galera-topology
+spec:
+  topologySpreadConstraints:
+  - maxSkew: 1
+    topologyKey: "topology.kubernetes.io/hostname"
+    whenUnsatisfiable: SchedulAnyway
```
dciabrin, line 16: Is the SchedulAnyway on purpose? I.e rather than ScheduleAnyway?

controllers/galera_controller.go
```diff
@@ -692,7 +706,45 @@ func (r *GaleraReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
...
+		ctx,
+		helper,
+		instance.Spec.TopologyRef,
+		&lastTopologyRef,
+		instance.Name,
+		mariadb.ServiceName,
+	)
+	if err != nil {
+		instance.Status.Conditions.Set(condition.FalseCondition(
+			condition.TopologyReadyCondition,
```
dciabrin, line 726: Do we need to reorder/amend the other Conditions checks in the existing unit tests?

fmount: No need to change any existing unit test because if `.Spec.TopologyRef` doesn't exist (which is the case of the tests already in place) the order is preserved. This provides fully backward compatibility and `TopologyCondition`  is not part of the default conditions (see L455), but only added if the current reconciliation has this parameter in its spec.
We can improve the newly introduced topology kuttl test (when I see a working execution) to assert all the conditions including Topology.

tests/kuttl/tests/galera_deploy_topology/01-deploy_galera_topology.yaml
fmount: usual typo :D fixing it!

fmount:
We need https://github.com/openstack-k8s-operators/install_yamls/pull/1015 to make topology CRD available in kuttl.

fmount:
Build step failed:
```
--> Creating resources with label build=openstack-operator ...
    error: the server is currently unable to handle the request
--> Failed
```
re-kicking kuttl.

stuggi (approved):
/lgtm

PR #316, opened 2025-03-31
don't assume any CSVs are installed

zzzeek:
change ae93738183eb86f950a62c6a77ab added the assumption that a CSV would be installed under
operators.coreos.com/openstack-operator.openstack-operators, but this cannot be assumed.  I run run-with-webhook with no csvs installed at all, so this check as well must only proceed with patch commands if CSV_NAME isn't blank.

### DIFF
--- a/hack/run_with_local_webhook.sh
+++ b/hack/run_with_local_webhook.sh
@@ -167,13 +167,15 @@ else
     # Handle operator deployed by Openstack Initialization resource
     CSV_NAME="$(oc get csv -n openstack-operators -l operators.coreos.com/openstack-operator.openstack-operators -o name)"
 
-    printf \
-    "\n\tNow patching openstack operator CSV to scale down deployment resource.
-    To restore it, use:
-    oc patch "${CSV_NAME}" -n openstack-operators --type=json -p=\"[{'op': 'replace', 'path': '/spec/install/spec/deployments/0/spec/replicas', 'value': 1}]\""
+    if [ -n "${CSV_NAME}" ]; then
+        printf \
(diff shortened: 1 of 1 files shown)


zzzeek:
just to clarify, the patch at ae93738183eb86f950a62c6a77ab  produces this output for me

```
++ oc get csv -n openstack-operators -l operators.coreos.com/mariadb-operator.openstack-operators -o name
+ CSV_NAME=
+ '[' -n '' ']'
++ oc get csv -n openstack-operators -l operators.coreos.com/openstack-operator.openstack-operators -o name
+ CSV_NAME=
+ printf '\n\tNow patching openstack operator CSV to scale down deployment resource.
    To restore it, use:
    oc patch  -n openstack-operators --type=json -p="[{'\''op'\'': '\''replace'\'', '\''path'\'': '\''/spec/install/spec/deployments/0/spec/replicas'\'', '\''value'\'': 1}]"'

        Now patching openstack operator CSV to scale down deployment resource.
    To restore it, use:
    oc patch  -n openstack-operators --type=json -p="[{'op': 'replace', 'path': '/spec/install/spec/deployments/0/spec/replicas', 'value': 1}]"+ oc patch '' -n openstack-operators --type=json '-p=[{'\''op'\'': '\''replace'\'', '\''path'\'': '\''/spec/install/spec/deployments/0/spec/replicas'\'', '\''value'\'': 0}]'
error: the server doesn't have a resource type ""
make: *** [Makefile:380: run-with-webhook] Error 1

```

karelyatin:
/lgtm
Thanks Michael, i had missed this case

PR #320, opened 2025-04-22
DNM: mariadbaccount system accounts

zzzeek:
This PR is broken into individual PRs:

#342 
#343 
#344 
#345 
#346 
#347

notes overall below:

this is for [OSPRH-14916](https://issues.redhat.com//browse/OSPRH-14916)

* reorganize StatefulSet/JobSpec use to get mysql root password from a cluster query script volume-mounted into containers, rather than embedding Secret / password in environment variables / CR variables which cause job hashes to change.   This will allow in-place changes to the name of the secret used for mysql root without kicking off jobs, and ensures the current root password can always be fetched appropriately in order to perform work on the database.
* introduce a new class of MariaDBAccount called a "system" MariaDBAccount, indicated by a new enumerated field AccountType on the CR.  Such accounts link directly to a Galera instance and have no dependency on a MariaDBDatabase CR.
* Ensure MariaDBAccount adds a finalizer to its own secret, and also cleans it out on delete
* solidify in-place password change ability by supporting finalizer remove for previous secret, improved messaging and account.sh operation
* change Galera CR to refer to a "root" MariaDBAccount object with its own secret, no longer using osp-secret.  The secret here will be generated in the usual way using EnsureMariaDBAccount before the Galera cluster is set up.

### DIFF
--- a/controllers/galera_controller.go
+++ b/controllers/galera_controller.go
@@ -33,7 +33,7 @@ import (
 	helper "github.com/openstack-k8s-operators/lib-common/modules/common/helper"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/labels"
 	common_rbac "github.com/openstack-k8s-operators/lib-common/modules/common/rbac"
-	secret "github.com/openstack-k8s-operators/lib-common/modules/common/secret"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/secret"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/service"
 	commonstatefulset "github.com/openstack-k8s-operators/lib-common/modules/common/statefulset"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/tls"
@@ -489,7 +489,7 @@ func (r *GaleraReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
 		{
 			APIGroups: []string{"mariadb.openstack.org"},
 			Resources: []string{"mariadbaccounts"},
-			Verbs:     []string{"get"},
+			Verbs:     []string{"get", "list"},
 		},
 		{
 			APIGroups: []string{""},
@@ -558,6 +558,8 @@ func (r *GaleraReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
 		log.Info("", "Kind", instance.Kind, "Name", instance.Name, "database service", service.Name, "operation", string(op))
 	}
 
+	instance.IsReady()
+
 	instance.Status.Conditions.MarkTrue(condition.CreateServiceReadyCondition, condition.CreateServiceReadyMessage)
 
 	// Map of all resources that may cause a rolling service restart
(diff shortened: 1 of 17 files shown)


zzzeek:
the last commit to add here will be to use MariaDBAccount for galera itself to generate / maintain the root password.

Just want to get some eyes on the various decisions being made here within each commit, particularly using an API call rather than env var for mariadb admin scripts to get galera root pw

zzzeek:
not related to the immediate failures but im noticing that deep in the prow scripts and probably elsewhere, osp-secret seems to be referenced for mariadb root password outside of the galera operator (that 12345678 password is no longer used as long as this change is kept generating a new root PW each time):

```
 rm -Rf /alabama/install_yamls/out/operator/nova-operator /alabama/install_yamls/out/mariadb-kuttl-tests/nova/cr
oc rsh openstack-galera-0 mysql -u root --password=12345678 -ss -e "show databases like 'nova_%';" | xargs -I '{}' oc rsh openstack-galera-0 mysql -u root --password=12345678 -ss -e "flush tables; drop database if exists {};"
Error from server (NotFound): pods "openstack-galera-0" not found
oc kustomize /alabama/install_yamls/out/mariadb-kuttl-tests/infra-redis/cr | oc delete --ignore-not-found=true -f -
error: must build at directory: not a valid directory: evalsymlink failure on '/alabama/install_yamls/out/mariadb-kuttl-tests/infra-redis/cr' : lstat /alabama/install_yamls/out/mariadb-kuttl-tests/infra-redis: no such file or directory
No resources found 
```

not totally sure how to fix this issue because this PR changes where the root password comes from.   only if I hardcoded it to look in "Galera->Secret->RootDbPassword" to get the initial password if a local mariadbaccount/secret doesnt exist yet.

zzzeek:
Well, what we can do is merge 9fff6b08199b81332c4bd5b6cadb18182f7d1bd1 ahead of everything else, then change all scripts everywhere to use the root_auth.sh script

so 

```
oc rsh openstack-galera-0 mysql -u root --password=12345678 -ss -e "show databases like 'nova_%';" | xargs -I '{}' oc rsh openstack-galera-0 mysql -u root --password=12345678 -ss -e "flush tables; drop database if exists {};"
```

becomes

```
oc rsh openstack-galera-0 <<EOF
#!/bin/bash
source /var/lib/operator-scripts/root_auth.sh

mysql -u root  -ss -e "show databases like 'nova_%';" | xargs -I '{}' mysql -u root  -ss -e "flush tables; drop database if exists {};"
EOF
```

that change can be made just with 9fff6b08199b81332c4bd5b6c merged alone as the controller that creates new galera pods, then all the scripts everywhere else can be abstracted away from where the galera root password comes from

zzzeek:
seeing as mariadb-operator-build-deploy-kuttl is now what's failing, I probably need to restore copying osp-secret as the initial root password just to get the code out the door as there are likely other parts of build / CI / deployment that are relying on this structure

controllers/galera_controller.go
```diff
@@ -552,8 +569,10 @@ func (r *GaleraReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
 	clusterPropertiesEnv := make(map[string]env.Setter)
 
 	// Check and hash inputs
-	// NOTE do not hash the db root password, as its change requires
-	// more orchestration than a simple rolling restart
+
+	// ******** TEMPORARY ************
+	// Pull the DbRootPassword from osp-secret to allow CI / external
+	// scripts to work, until they can adapt to root_auth.sh
```
zzzeek, line 575: some view into what scripts need modification can be seen using this query:

https://github.com/search?q=org%3Aopenstack-k8s-operators+DbRootPassword&type=code&p=1

dciabrin:
I needed to push an update to the galera CRD not so long ago, so this needs to be rebased...

I wonder if we could take that opportunity to maybe split those commits into separate PR (as suggested by @zzzeek to help move this forward), so that we can do a simpler review before merging?

I had a chat with @zzzeek already to discuss the details behind the "system" MariaDBAccount approach, and I think it's solid and that's where we want to go. We now need to carefully review those changes before merging.

zzzeek:
OK do you want all six ?   if that makes it easier

zzzeek:
six it is

PR #339, opened 2025-06-25
Expose pod property for resource QoS

dciabrin:
Add the ability to configure required/allowed CPU and memory consumption for galera pod, by exposing the standard QoS Resources from pod API.

The galera type now exposes an optional Resources field to configure galera pod's QoS, similarly to the RabbitMQCluster type. Unlike rabbitmq, no default QoS is enforce, so the default galera pod behaviour doesn't change (QoS Class is BestEffort).

Jira: [OSPRH-17410](https://issues.redhat.com//browse/OSPRH-17410)


stuggi (commented):
/lgtm

PR #342, opened 2025-07-01
Get MYSQL_PWD using an on-demand cluster query (PR 1 of 6)

zzzeek:
In order to facilitate an in-place change to the name of the Secret that is referenced by a Galera instance for the mysql root password, rework
the approach used by pods and shell scripts to no longer require the root secret name and/or password be passed by environment variable, instead using a pod-level cluster query to retrieve the current root password.  The logic to retrieve this password is encapsulated into a single shell script that is present as a volume mount on running containers.

This allows Job objects to be created with hashes that do not link to a specific Secret name, as well as to create StatefulSet objects that don't refer to this name. When the Secret name changes on a Galera instance for an in-place root password change, the hashes / CRs for these objects will remain unchanged.

A subsequent change to the mariadb operator will add the ability to change the mysql root password of a Galera cluster using a dual-reference architecture where
the "current" root secret will be part of <CR>/Status, while the secret referenced in <CR>/Spec will be the "new" root secret.  When these two names differ, that will indicate an in-place password change should take place, as well as allowing the pre-existing root password to be available at the same time as the new one in order to do a root password
change.   The same
architecture will be applied to a new class of "system" MariaDBAccount
objects that are for use only by the Galera instance itself
and do not have a link to any MariaDBDatabase CR.  The
Galera CR itself will no longer use osp-secret
for the mysql root password nor will the secret be directly
referenced from the Galera CR, instead referenced by a
"system" MariaDBAccount CR which the Galera operator itself
will create.


controllers/galera_controller.go
```diff
@@ -486,6 +486,21 @@ func (r *GaleraReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
 			Resources: []string{"services"},
 			Verbs:     []string{"get", "list", "update", "patch"},
 		},
+		{
+			APIGroups: []string{"mariadb.openstack.org"},
+			Resources: []string{"galeras"},
+			Verbs:     []string{"get", "list"},
+		},
+		{
+			APIGroups: []string{"mariadb.openstack.org"},
```
dciabrin, line 495: We don't need that specific RBAC in this PR

pkg/mariadb/volumes.go
```diff
@@ -117,6 +103,10 @@ func getGaleraVolumes(g *mariadbv1.Galera) []corev1.Volume {
 							Key:  "mysql_wsrep_notify.sh",
 							Path: "mysql_wsrep_notify.sh",
 						},
+						{
+							Key:  "root_auth.sh",
```
dciabrin, line 107: Nitpick: should we call it mysql_root_auth.sh to avoid any misconception with UNIX root?

templates/galera/bin/detect_last_commit.sh
```diff
@@ -2,6 +2,9 @@
 
 set -eu
 
+source /var/lib/operator-scripts/root_auth.sh
+
+
```
dciabrin, line 7: Only one blank line to match with other updates

tests/chainsaw/scripts/mysql-cli.sh
```diff
@@ -2,5 +2,7 @@
 set -u
 ARGS=$1
 SQL=$2
+ROOT_PW_AUTH="source /var/lib/operator-scripts/root_auth.sh; "
```
dciabrin, line 5: We don't need the space after the semicolon

templates/galera/bin/root_auth.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+set +eux
```
dciabrin, line 2: Do not use +x here otherwise when mysql probes fail the entire debug output gets logged in k8s and shows up in `oc get events`.

templates/galera/bin/root_auth.sh
```diff
@@ -0,0 +1,39 @@
...
+SERVICEACCOUNT=/var/run/secrets/kubernetes.io/serviceaccount
+NAMESPACE=$(cat ${SERVICEACCOUNT}/namespace)
+TOKEN=$(cat ${SERVICEACCOUNT}/token)
+CACERT=${SERVICEACCOUNT}/ca.crt
+K8S_API="api/v1"
+MARIADB_API="apis/mariadb.openstack.org/v1beta1"
+
+GALERA_INSTANCE="{{.galeraInstanceName}}"
+
+# note jq is not installed in the galera image, macgyvering w/ python instead
```
dciabrin, line 18: Ack, I'll track the shortcoming in a separate Jira so we can fix that for good.

dciabrin (commented):
Thanks for that.
The only problem I have with this review is that we currently use the db root user for probes, because we haven't got yet a clean way of creating alternative users (part of it will be resolved with your next review I think). So if we were to merge it as is, we'd end up doing two API calls for each probe sent to the galera pods, which is way too many.

I think we should discuss how to best cache that and only call the API when needed. I was thinking something along the lines of:
  1. we keep using root_auth any time we want to get root creds.
  2. internally, root_auth checks whether /root/.my.cnf exists. If it doesn't, it calls the API and generate that file with the current password.
  3. if it exists, it checks whether creds are still valid (e.g. by doing a mysqladmin ping). If not, do 2.
  4. if mysql is not running (i.e. when pod start), do not try to check creds validity, assume .my.cnf is valid and wait for the next probe to update it if needed.

PR #345, opened 2025-07-01
mariadbaccount system accounts (PR 4 of 6)

zzzeek:
introduce a new class of MariaDBAccount called a "system"
MariaDBAccount, indicated by a new enumerated field AccountType
on the CR.  Such accounts link directly to a Galera instance
and have no dependency on a MariaDBDatabase CR.

The expected targets for "system" accounts will include the
Galera/mysql root username and password, as well as a system
account used by mariadbbackup for SST.

Refactor mariadbaccount_controller to isolate logic used for
acquiring MariaDBDatabase and Galera CRs into separate functions,
and ensure all MariaDBDatabase logic takes place only for "user"
accounts (which would be all current MariaDBAccount CRs).

Also correct an oversight where MariaDBAccount would not unconditionally
apply a finalizer to its Secret object.   This logic now takes place
in addition to an unconditional removal of the finalizer when the
MariaDBAccount object is deleted.

A subsequent change will allow system-level passwords to be changed
in place by applying the secret name to two separate fields
MariaDBAccount/Spec/Secret and MariaDBAccount/Status/Secret.  When
these two names differ it will indicate an in-place password change
should take place.

### DIFF
--- a/controllers/mariadbaccount_controller.go
+++ b/controllers/mariadbaccount_controller.go
@@ -30,6 +30,7 @@ import (
 	util "github.com/openstack-k8s-operators/lib-common/modules/common/util"
 	databasev1beta1 "github.com/openstack-k8s-operators/mariadb-operator/api/v1beta1"
 	mariadb "github.com/openstack-k8s-operators/mariadb-operator/pkg/mariadb"
+	batchv1 "k8s.io/api/batch/v1"
 	k8s_errors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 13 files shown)


controllers/mariadbaccount_controller.go
```diff
@@ -112,13 +113,23 @@ func (r *MariaDBAccountReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
-	cl := condition.CreateList(
-		condition.UnknownCondition(condition.ReadyCondition, condition.InitReason, condition.ReadyInitMessage),
-		condition.UnknownCondition(databasev1beta1.MariaDBServerReadyCondition, condition.InitReason, databasev1beta1.MariaDBServerReadyInitMessage),
-		condition.UnknownCondition(databasev1beta1.MariaDBDatabaseReadyCondition, condition.InitReason, databasev1beta1.MariaDBDatabaseReadyInitMessage),
-		condition.UnknownCondition(databasev1beta1.MariaDBAccountReadyCondition, condition.InitReason, databasev1beta1.MariaDBAccountReadyInitMessage),
-	)
+	var cl condition.Conditions
+	if instance.IsUserAccount() {
+		// initialize conditions used later as Status=Unknown
+		cl = condition.CreateList(
```
dciabrin: Is there a means to benefit from the variadic signature of function condition.CreateList to pass the condition for MariaDBDatabaseReadyCondition conditionally instead of duplicating the lines 120..130?

controllers/mariadbaccount_controller.go
```diff
@@ -300,13 +338,24 @@ func (r *MariaDBAccountReconciler) reconcileDelete(
...
+	if instance.IsUserAccount() {
+
+		log.Info(fmt.Sprintf("Running account delete '%s' MariaDBDatabase '%s'", instance.Name, mariadbDatabase.Spec.Name))
+
+		jobDef, err = mariadb.DeleteDbAccountJob(dbGalera, instance, mariadbDatabase.Spec.Name, dbHostname, dbContainerImage, serviceAccountName, dbGalera.Spec.NodeSelector)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	} else {
+		log.Info(fmt.Sprintf("Running system account delete '%s'", instance.Name))
```
dciabrin: We likely want to enforce that the root user is never deleted from the database, otherwise we won't be able to recreate any new one afterwards.
We need a way to prevent deletion of such user, unless you don't plan on mapping this root user to a System MariaDBAccount?

dciabrin (changes requested):
I need to understand how the deletion of a MariaDBAccount of type System would work if we plan to create one for the root database user. Right now I have the feeling we should have a safeguard mechanism to prevent deleting root.

Also, could you please rewrite the new KUTTL test into a Chainsaw test? I'm trying to get rid of all remaining KUTTL tests before disabling that CI test in prow to free up some CI resource.

controllers/mariadbaccount_controller.go
zzzeek: this is necessarily a bit awkward because the "user" element is in the middle of the list, and we don't want to have to change the order of the elements in all the assertion YAMLs (plus I dont know if there are other such yamls external to this repo).

PR #346, opened 2025-07-01
support in-place secret changes (PR 5 of 6)

zzzeek:
The existing mariadb operator in fact already "supports" in-place
change of secret, since if you change Spec.Secret to a new secret
name, that would imply a new job hash, and the account.sh script running
using only GRANT statements would update the password per mariadb.
This already works for flipping the TLS flag on and off too.

So in this patch, we clean this up and add a test to include:

* a new field Status.CurrentSecret, which is used to indicate the
  previous secret from which the finalizer should be removed.  This will
  also be used when we migrate the root password to use MariaDBAccount
  by providing the "current" root password when changing to a new root
  password
* improved messaging in log messages, name of job.   This changes the
  job hash for mariadbaccount which will incur a run on existing environments,
  however the
  job hashes are already changing on existing environments due to the
  change in how the mysql root password is sent, i.e. via volume mounted
  script rather than env var secret
* update account.sh to use modern idiomatic patterns for user create
  /alter, while mariadb is fine with the legacy style of using only
  GRANT statements, MySQL 8 no longer allows this statement to proceed
  without a CREATE USER, so formalize the commands used here to use
  distinct CREATE USER, ALTER USER, GRANT statements and clarify the
  script is good for all create/update user operations.

### DIFF
--- a/controllers/mariadbaccount_controller.go
+++ b/controllers/mariadbaccount_controller.go
@@ -129,17 +129,17 @@ func (r *MariaDBAccountReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	instance.Status.Conditions.Init(&cl)
 
 	if instance.DeletionTimestamp.IsZero() || isNewInstance { //revive:disable:indent-error-flow
-		return r.reconcileCreate(ctx, log, helper, instance)
+		return r.reconcileCreateOrUpdate(ctx, log, helper, instance, isNewInstance)
 	} else {
 		return r.reconcileDelete(ctx, log, helper, instance)
 	}
 
 }
 
-// reconcileDelete - run reconcile for case where delete timestamp is zero
-func (r *MariaDBAccountReconciler) reconcileCreate(
+// reconcileCreateOrUpdate - run reconcile for case where delete timestamp is zero
+func (r *MariaDBAccountReconciler) reconcileCreateOrUpdate(
 	ctx context.Context, log logr.Logger,
-	helper *helper.Helper, instance *databasev1beta1.MariaDBAccount) (result ctrl.Result, _err error) {
+	helper *helper.Helper, instance *databasev1beta1.MariaDBAccount, isNewInstance bool) (result ctrl.Result, _err error) {
 
 	var mariadbDatabase *databasev1beta1.MariaDBDatabase
 	var err error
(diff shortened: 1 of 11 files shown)


dciabrin (changes requested):

controllers/mariadbaccount_controller.go
```diff
@@ -226,11 +235,24 @@ func (r *MariaDBAccountReconciler) reconcileCreate(
...
+
 		if instance.Status.Hash == nil {
 			instance.Status.Hash = make(map[string]string)
 		}
 		instance.Status.Hash[databasev1beta1.AccountCreateHash] = accountCreateJob.GetHash()
-		log.Info(fmt.Sprintf("Job %s hash added - %s", jobDef.Name, instance.Status.Hash[databasev1beta1.AccountCreateHash]))
+		log.Info(fmt.Sprintf("Job %s hash created or updated - %s", jobDef.Name, instance.Status.Hash[databasev1beta1.AccountCreateHash]))
+
+		// set up new Secret and remove finalizer from old secret
+		if instance.Status.CurrentSecret != instance.Spec.Secret {
```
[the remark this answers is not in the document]
dciabrin: Oh right, I misinterpreted the semantics of HasChanged().

controllers/mariadbaccount_controller.go
dciabrin: I just notice that to give context you renamed CreateDB to CreateOrUpdateDb (+1 for that), but you did not do it for the accountCreateHash variable. So the intent of this variable is a bit lost there. We should probably update it to read CreateOrUpdate? Same for accountCreatJob below.

controllers/mariadbaccount_controller.go
```diff
@@ -226,11 +235,24 @@ func (r *MariaDBAccountReconciler) reconcileCreate(
 	}
 
 	if accountCreateJob.HasChanged() {
+
 		if instance.Status.Hash == nil {
 			instance.Status.Hash = make(map[string]string)
 		}
 		instance.Status.Hash[databasev1beta1.AccountCreateHash] = accountCreateJob.GetHash()
-		log.Info(fmt.Sprintf("Job %s hash added - %s", jobDef.Name, instance.Status.Hash[databasev1beta1.AccountCreateHash]))
+		log.Info(fmt.Sprintf("Job %s hash created or updated - %s", jobDef.Name, instance.Status.Hash[databasev1beta1.AccountCreateHash]))
```
dciabrin: I'm not sure we should replace the word added here. It seems like this was meant to signal that the Job CR finished its run.
If we change that log it now reads the hash itself was created or updated, which is slightly misleading, and it likely differs from all other logs in this repo where we need to run Jobs and log their state.

PR #348, opened 2025-07-18
Fetch up-to-date gcomm members list during a failover

dciabrin:
The operator script that implements service endpoint failover
contains internal logic to probe the up-to-date state of the
gcomm cluster. This is done when the script starts, or when
a command failed and is retried.

The list of members was incorrectly extracted from a mysql
table which is not guaranteed to be up-to-date when e.g.
a node disappears from the cluster due to a network partition.

Instead, we must rely on the mysql status, that always exposes
the up-to-date gcomm state, in particular the members that
are still connected to the primary partition.

Jira: [OSPRH-18408](https://issues.redhat.com//browse/OSPRH-18408)


dciabrin:
/retest
unrelated failure prior to tests in ci/prow/mariadb-operator-build-deploy

PR #436, opened 2026-03-30
improve fallback PW check for upgrades, harden against broken installs

zzzeek:
testing has shown it's possible to deploy with pre-FR5 versions with passwords that interfere with bootstrap, leading
to the "root@%" account missing entirely.   Ensure bootstrap
makes sure this is there for recovery scenarios.  next, an FR4
deploy may have the "rootDatabaseAccount" key omitted entirely from the
CR; make the python fetch more robust and add a test that simulates
the full scenario that was observed.

### DIFF
--- a/test/chainsaw/tests/root-pw-fallback/chainsaw-test.yaml
+++ b/test/chainsaw/tests/root-pw-fallback/chainsaw-test.yaml
@@ -1,0 +1,148 @@
+apiVersion: chainsaw.kyverno.io/v1alpha1
+kind: Test
+metadata:
+  name: root-pw-fallback
+spec:
+  steps:
+  - name: Deploy 3-node cluster
+    description: Deploy a 3-node galera cluster
+    bindings:
+    - name: replicas
+      value: 3
+    try:
+    - apply:
+        file: ../../common/galera.yaml
+    - assert:
+        file: ../../common/galera-assert.yaml
+
+  - name: Verify initial root account exists
+    description: Verify the root MariaDBAccount exists with initial secret
+    try:
+    - assert:
+        file: root-account-initial-assert.yaml
+
+  - name: Capture original password and delete root at percent
+    description: >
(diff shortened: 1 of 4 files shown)


dciabrin (changes requested):
I'd prefer to rely on chainsaw's "outputs" functionality to simplify the test.
Also the original commit message should be updated, it has a incomplete sentence.

test/chainsaw/tests/root-pw-fallback/chainsaw-test.yaml
```diff
@@ -0,0 +1,157 @@
...
+          oc exec -n ${NAMESPACE} -c galera openstack-galera-0 -- /bin/sh -c "
+            mysql -uroot -p'${ORIG_PW}' -e \"DROP USER 'root'@'%'; FLUSH PRIVILEGES;\"
+          "
+          echo "Dropped root@% user (replicates to all nodes)"
+    - script:
+        env:
+        - name: ORIG_PW
+          value: ($orig_pw)
+        content: |
+          sleep 3
```
dciabrin: I don't think we should keep this sleep, it seems a bit arbitrary to me. The SQL statement is synchronously replicated so at least it's not needed for that. Plus you're reconnecting to the same node, so you're guaranteed to have the new credentials anyway.

dciabrin (changes requested):
Small nitpicks, but let's fix that now so we're covered for future updates.

test/chainsaw/tests/root-pw-fallback/chainsaw-test.yaml
zzzeek: I dunno the test was not passing, then i put in that sleep, then it was fine.  my guess was that the "ALTER USER" statement hadnt replicated to all three nodes, or something.  I can try again.

dciabrin (approved):

PR #450, opened 2026-04-24
Increase hard-coded liveness and readiness probe timeouts

dciabrin:
Liveness and readiness probes currently rely on the default timeout of 1s, which is low and might cause unecessary instabilities on loaded environment.
Set an explicit default of 5s, until we move to configurable timeouts with the help of lib-common's functions.


dciabrin:
Closed in favour of #452

PR #452, opened 2026-04-29
Add configurable probe timeouts to Galera CR

zzzeek:
use lib-common ProbeConf as a starting base for Probe values, then apply values from the Galera.Spec into each ProbeConf. then create Probe objects from that.

The lib-common CreateProbeSet function is hardcoded to HTTP probes, so we can't use the new features fully, so here we vendor our own createProbeFromConf function.  we also create the ProbeConf with user-merged values up front so we have a mergeProbeConf function
that resembles lib-common.probe.merge().   lib-common's library could
have some more public API features added to address these cases.

patch also includes validating webhooks w/ unit tests as well as a chainsaw test modeled after the qos test suite.

References: [OSPRH-29441](https://redhat.atlassian.net/browse/OSPRH-29441)


config/crd/bases/mariadb.openstack.org_galeras.yaml
```diff
@@ -150,6 +196,29 @@ spec:
...
+                properties:
+                  failureThreshold:
+                    format: int32
+                    minimum: 1
+                    type: integer
+                  initialDelaySeconds:
+                    format: int32
+                    minimum: 0
+                    type: integer
+                  path:
```
dciabrin: IIUC those new fields are the copy of what's in libcommon right now, because the exposed API doesn't fit us due to the ExecAction that we need.
Then if we duplicate, do we really need to keep the `path` field?

dciabrin (commented):
Thanks Mike.

So right now we're not in immediate need of this improvement, so I'm wondering if we should push for a more generic interface in libcommon first, like it is done in https://pkg.go.dev/k8s.io/api/core/v1#ProbeHandler

This way we'd have a single review to push in mariadb-operator instead of this one plus an improved one once we agree on the libcommon update.

@fmount any opinion?

fmount:
> Thanks Mike.
> 
> So right now we're not in immediate need of this improvement, so I'm wondering if we should push for a more generic interface in libcommon first, like it is done in https://pkg.go.dev/k8s.io/api/core/v1#ProbeHandler
> 
> This way we'd have a single review to push in mariadb-operator instead of this one plus an improved one once we agree on the libcommon update.
> 
> @fmount any opinion?

Hi @dciabrin @zzzeek, 
I agree that if this is not something we need to backport to a previous FR we could propose an enhancement to `lib-common` (e.g. we can export the functions we need here and add any missing piece), bump the dependency and then finalize this patch. I think extending a common interface to support multiple use cases would be the best option.

config/crd/bases/mariadb.openstack.org_galeras.yaml
zzzeek: OK just to confirm, the suggestion at the top of the PR is to enhance lib-common ProbeConf to include ExecAction as well, so we'd have even "*more* fields here (which we dont technically need since our execaction is fixed)?

or are we looking to make a new structure like ProbeTimingConf that can be used in CRs to only expose timeouts?

fmount: @zzzeek I would try to add the missing field because there's no distinction between `ProbeTimingConf` and `ProbeHandlerConf` in the lib-common struct. We could do this distinction but it should be transparent to the operators that already adopted this interface to override probes.
It's true that you would get even more params here, but it's hard to find a set of fields that are valid across the operators due to the different needs.

fmount: As far as I understand the gap is narrow; `lib-common` just needs two things to cover this use case:

- Export `merge`, which seems a trivial change
- Add support for `Exec` probe handlers (either a `SetExecProbeConf` function, or a `ProbeHandlerType` parameter on the existing functions is good).

If you like to have a `ProbeHandlerType` is good too, and should work transparently for other operators as well.

zzzeek: OK that's fine, I just want to show you what me and the robot came up with, which would change the API but the idea is that the "timing" part would be something downstream operators can put into their CRs the way this PR is doing with ProbeConf, that would not expose the path/execcommand part that is typically going to be hardcoded

```
// NEW in lib-common
type ProbeTiming struct {
    // +kubebuilder:validation:Minimum=0
    InitialDelaySeconds int32 `json:"initialDelaySeconds,omitempty"`
    // +kubebuilder:validation:Minimum=1
    TimeoutSeconds      int32 `json:"timeoutSeconds,omitempty"`
    // +kubebuilder:validation:Minimum=1
    PeriodSeconds       int32 `json:"periodSeconds,omitempty"`
    // +kubebuilder:validation:Minimum=1
    FailureThreshold    int32 `json:"failureThreshold,omitempty"`
}

// MODIFIED ProbeConf
type ProbeConf struct {
    ProbeTiming `json:",inline"`  // Embeds all timing fields at same level
    // +kubebuilder:validation:Pattern=`^(/.*)?$`
    Path        string   `json:"path,omitempty"`
    Command     []string `json:"command,omitempty"`
}
```

but this is kind of just cosmetic stuff.   I think though that exposing "path" and "command" in CRs is mostly not very useful though IIUC since that part is going to be application-hardcoded

fmount: In general in the operators we have a set of default probes, and the proposal of the lib-common module is to provide an interface to override the behavior. We're not going to expose `ProbeTiming` as part of the Spec
(see [1] and [2] to get more context). Probes wasn't something we wanted to expose as part of the Spec, but a potential override to help us dealing w/ customer cases better, so we exposed only the parameters we need deal with.

> but this is kind of just cosmetic stuff. I think though that exposing "path" and "command" in CRs is mostly not very useful though IIUC since that part is going to be application-hardcoded

It depends, we have places where we don't care, other parts where it makes sense, so I think the deal is that if an operator doesn't need it, it can simply be ignored (not set).
The change you're proposing seems fine and reusable, though I think we need a small follow up in lib-common to extend the internals (and tests), but we can discuss this on that patch. 


[1] https://github.com/openstack-k8s-operators/cinder-operator/blob/main/docs/probes.md
[2] https://github.com/openstack-k8s-operators/cinder-operator/blob/main/internal/cinder/funcs.go#L93

zzzeek: OK so add "Command" to ProbeConf and... that's all we want then, it looks like Cinder is also exposing ProbeConf in its CR and has "path" there, even though it's ignored later.    


@dciabrin should we mimic the "override" pattern that Cinder uses?   like this

```
// In galera_types.go
type GaleraSpecCore struct {
    // ... existing fields
    Resources corev1.ResourceRequirements `json:"resources,omitempty"`
    
    // +kubebuilder:validation:Optional
    // Override, provides the ability to override the generated manifest
    Override GaleraOverrideSpec `json:"override,omitempty"`
}

type GaleraOverrideSpec struct {
    // Override probes and other common fields in the StatefulSet
    Probes probes.OverrideSpec `json:"probes,omitempty"`
}
```

this would present the fields in the CR underneath "overrides".   we dont seem to have any fields like that now

dciabrin: It looks like Cinder and Glance are already going that path, I'm good with doing the same (maybe eventually we should move `CustomServiceConfig` there as well...).

fmount: manila adopted the same interface/pattern for probes as well [1].

[1] https://github.com/openstack-k8s-operators/manila-operator/blob/main/api/v1beta1/manilaapi_types.go#L68

zzzeek: > As far as I understand the gap is narrow; `lib-common` just needs two things to cover this use case:
> 
>     * Export `merge`, which seems a trivial change
> 
>     * Add support for `Exec` probe handlers (either a `SetExecProbeConf` function, or a `ProbeHandlerType` parameter on the existing functions is good).
> 
> 
> If you like to have a `ProbeHandlerType` is good too, and should work transparently for other operators as well.

So yes, I do want to do ProbeHandlerType, and I want to leave SetProbeConf, more or less.

My only thing is if we are having SetProbeConf (or whatever it's called) look at the ProbeConf.ProbeType to determine what fields we care about, do we want to have *all* the fields on ProbeConf so that it can switch on type?  I'm talking specifically about "port" and "scheme" right now which appear to be sent separately to SetProbeConf.

The most simple change that I can just start with is:

1. create ProbeHandlerType, which has ProbeHandlerHTTP, ProbeHandlerExec members
2. add ProbeConf->Type which is a ProbeHandlerType (and defaults to HTTP if the value is "" since that's what's currently out there)
3. add ProbeConf->Command (String [])
4. change SetProbeConf internally to look at this type, and if it's Exec it will **ignore** the "port" and "scheme" parameters, if it's HTTP will make use of these params (similar for ValidateProbeConf)

this has no backwards-incompatibility and keeps the whole HTTP version of things working exactly the same.

The additional optional step is to add port/scheme to ProbeConf and deprecate their use on SetProbeConf

fmount: +1 to extend the struct: given that we `inline` the existing fields we don't have to rebuild the CRD. This way we can go through steps 1-3 without big problems or without requiring any change to the existing operators.
Regarding step 4, changing the internal logic would be one thing, and we could keep the signature unchanged.
However, if we want to make the signature better, for example:

```golang
 func SetProbeConfV2(config ProbeConf) (*v1.Probe, error) {
      probe := &v1.Probe{
          InitialDelaySeconds: config.InitialDelaySeconds,
          TimeoutSeconds:      config.TimeoutSeconds,
          PeriodSeconds:       config.PeriodSeconds,
          FailureThreshold:    config.FailureThreshold,
      }

      switch config.Type {
      case ProbeHandlerExec:
      ....
      ....
```
we could keep `SetProbeConf` as it is, and introduce a `SetProbeConfV2` like the above. `mariadb-operator` can start using it, and after we bump `lib-common` across the board we can push an update to the operators.
It's just an option/strategy, but I assume we can start a lib-common patch and we can discuss all the details there.
(FYI @stuggi)

zzzeek: proposed in https://github.com/openstack-k8s-operators/lib-common/pull/696

dciabrin:
chainsaw CI test fixes just landed, so since the lib-common part has lnaded, I think we can move forward with this one now.

zzzeek:
need to update go.mod appropriately

api/go.sum
```diff
@@ -133,6 +131,8 @@ github.com/x448/float16 v0.8.4 h1:qLwI1I70+NjRFUR3zs1JPUCgaCXSh3SW62uAKT1mSBM=
 github.com/x448/float16 v0.8.4/go.mod h1:14CWIYCyZA/cWjXOioeEpHeN/83MdbZDRQHoFcYsOfg=
 github.com/yuin/goldmark v1.1.27/go.mod h1:3hX8gzYuyVAZsxl0MRgGTJEmQBFcNTphYh9decYSb74=
 github.com/yuin/goldmark v1.2.1/go.mod h1:3hX8gzYuyVAZsxl0MRgGTJEmQBFcNTphYh9decYSb74=
+github.com/zzzeek/lib-common/modules/common v0.0.0-20260519194359-f7a36c578816 h1:aYBfRbS/nPXy6+KnRTnaEnxpVhid/G0pe9qOwAVjEvY=
```
zzzeek: hrm, i thought i updated this....moment

zzzeek:
let's try a rebase to get those recent chainsaw fixes

dciabrin:
https://github.com/openstack-k8s-operators/mariadb-operator/pull/474 aiming at improving the stability of chainsaw tests landed.

dciabrin (commented):
LGTM

dciabrin:
@zzzeek can you rebase the PR so it picks up the change for chainsaw?

zzzeek:
/retest

that one was the pool

dciabrin (approved):

PR #460, opened 2026-05-16
Truncate backup CronJob name to avoid exceeding K8s label limit

lmiccini:
When the combined CronJob name (galera + backup CR names) exceeds 47 characters, the auto-generated Job name with its timestamp suffix can exceed the 63-character Kubernetes label value limit, causing Job creation to fail. Truncate and append a short SHA256 hash to keep names unique while staying within the limit.

### DIFF
--- a/internal/mariadb/backup/utils.go
+++ b/internal/mariadb/backup/utils.go
@@ -1,10 +1,37 @@
 package mariadbbackup
 
-import mariadbv1 "github.com/openstack-k8s-operators/mariadb-operator/api/v1beta1"
+import (
+	"crypto/sha256"
+	"fmt"
 
-// BackupCronJobName - name of the cronjob resource created for a backup CR
+	mariadbv1 "github.com/openstack-k8s-operators/mariadb-operator/api/v1beta1"
+)
+
+const (
+	// backupJobMaxLabelLen is the maximum CronJob name length that keeps
+	// auto-generated Job names within the 63-character Kubernetes label
+	// value limit. We reserve 16 characters for the timestamp suffix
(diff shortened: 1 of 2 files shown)


stuggi:
wondering if it would be just easier to call the cronjob like the galerabackup and just not do it `<galerabackup>-backup-<galeraname>`. we already have all the data as labels in the cronjob:

```
  labels:                           
    app: galera                             
    cr: galera-openstack                                                                      
    galera/name: openstack                
    galera/namespace: openstack   
    galera/uid: ce268643-f3d2-4418-a1e1-e353baea9f51            
    owner: mariadb-operator     
    service: openstack-galera    
```

stuggi:
In any case I think we have to delete the old one and re-create it for updates? otherwise we end up with 2 cronjobs per galerabackup?

dciabrin:
The idea behind the naming convention for the resources created by the GaleraBackup CR was to allow different backups on the same galera CR, as well as to avoid ambiguities when listing PVC/cronjob and whatnot.

If we start hashing for reclaiming size, I'm concerned we'll end up making day two operation more tedious, so is this issue theoretical or is it already impacting test deployments?

If anything, I'd prefer to remove ref to the galera DB in the name to shorten the generated identifier rather than obfuscate the label with a hash. I can perform internal tests and report back.

lmiccini:
so we drop this for https://github.com/openstack-k8s-operators/mariadb-operator/pull/463 ?

lmiccini:
going with https://github.com/openstack-k8s-operators/mariadb-operator/pull/463

PR #463, opened 2026-05-18
Shorten name of resources created by galerabackup

dciabrin:
This prevents cases where instantiating job would fail when labels configured to set up the job exceed the k8s limit of 63 characters.

Note: by changing the naming convention of the pvc and cronjob generated for a galerabackup CR, we are leaving behind old resources.

### DIFF
--- a/internal/mariadb/backup/utils.go
+++ b/internal/mariadb/backup/utils.go
@@ -1,23 +1,31 @@
 package mariadbbackup
 
-import mariadbv1 "github.com/openstack-k8s-operators/mariadb-operator/api/v1beta1"
+import (
+	labels "github.com/openstack-k8s-operators/lib-common/modules/common/labels"
+	mariadbv1 "github.com/openstack-k8s-operators/mariadb-operator/api/v1beta1"
+)
 
 // BackupCronJobName - name of the cronjob resource created for a backup CR
-func BackupCronJobName(b *mariadbv1.GaleraBackup, g *mariadbv1.Galera) string {
-	return g.Name + "-backup-" + b.Name
+func BackupCronJobName(b *mariadbv1.GaleraBackup) string {
+	return "backup-" + b.Name
 }
 
 // BackupPVCName returns the name of the PVC resource that stores .sql backup files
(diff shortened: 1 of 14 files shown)


internal/mariadb/backup/utils.go
```diff
@@ -4,20 +4,20 @@ import mariadbv1 "github.com/openstack-k8s-operators/mariadb-operator/api/v1beta
...
 // BackupTransferPVCName returns the name of the PVC resource that stores temporary binary backup data
 func BackupTransferPVCName(b *mariadbv1.GaleraBackup, g *mariadbv1.Galera) string {
-	return "mysql-transfer-" + g.Name + "-backup-" + b.Name
+	return "mysql-transfer-" + b.Name
 }
 
 // RestorePodName - name of the pod resource created by a GaleraRestore CR
 func RestorePodName(r *mariadbv1.GaleraRestore, g *mariadbv1.Galera) string {
-	return g.Name + "-restore-" + r.Name
+	return "restore-" + r.Name
```
stuggi, line 25: maybe in https://github.com/openstack-k8s-operators/mariadb-operator/blob/main/internal/mariadb/backup/restorepod.go#L13 we can add a label to the pod with the galerarestore name? for documentation on which pod to wait we could remove the need to know the naming format. just for making https://github.com/openstack-k8s-operators/ci-framework/pull/3886/changes#diff-10e89b09a7eb086f3741d3e56cf2cf72f02df2a9c88f0ef89db04e8585c9673bR331 and https://github.com/openstack-k8s-operators/dev-docs/blob/main/backup-restore/user-guide.md#step-7-service-database-restore a bit more generic.

stuggi: or we just document that its `restore-<galerarestore name>`.

dciabrin: Something like that? https://github.com/openstack-k8s-operators/mariadb-operator/pull/463/changes#diff-0a7b4acada262caa77c6d3ab82eabc96a44e4a3e3f3041cb19227e11a4218220R19

abays (commented):
+1 from me

stuggi (approved):
/lgtm

PR #467, opened 2026-05-23
Fix missing-dbrootpassword chainsaw test failing in CI

lmiccini:
Remove hardcoded containerImage from the assert file. The test applies a Galera CR without specifying containerImage, so the webhook defaults it. In CI the default is a sha256 digest, which never matches the hardcoded tag.


lmiccini:
this should fix the issue seen in https://github.com/openshift/release/pull/79648

zzzeek (approved):
claude FTW on that.   we have to figure out how to catch it doing things like this

PR #468, opened 2026-05-23
Fix restore chainsaw test race condition on pod readiness

lmiccini:
The chainsaw restore test was failing because it tried to exec into restore pods immediately after the GaleraRestore CR became Ready, but the CR is marked Ready as soon as the pod is created, not when the container is actually running. This adds a step to assert pod container readiness before exec'ing into them.

### DIFF
--- a/test/chainsaw/tests/restore/chainsaw-test.yaml
+++ b/test/chainsaw/tests/restore/chainsaw-test.yaml
@@ -71,6 +71,31 @@ spec:
     - assert:
         file: cluster2restore-assert.yaml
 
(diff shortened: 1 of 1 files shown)


lmiccini:
seen in https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/79648/rehearse-79648-pull-ci-openstack-k8s-operators-mariadb-operator-main-mariadb-operator-build-deploy-chainsaw/2058207686742249472

dciabrin (approved):

PR #469, opened 2026-05-25
Fix restore chainsaw test asserting wrong cluster name

lmiccini:
The cluster2-1node-assert.yaml was asserting on name: openstack instead of name: cluster2, so the test never waited for the cluster2 Galera cluster to be ready. This caused a race condition where the next step tried to exec into cluster2-galera-0 before MySQL was accepting connections, resulting in socket connection errors.


lmiccini:
seen in https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/79648/rehearse-79648-pull-ci-openstack-k8s-operators-mariadb-operator-main-mariadb-operator-build-deploy-chainsaw/2058780856641654784

dciabrin (approved):

PR #472, opened 2026-05-26
Let pods report their state to the operator

dciabrin:
Reverse the introspection approach for bootstrapping the cluster. Until now, the mariadb operator was in charge in running scripts inside pods to extract the state of the mysql database.

Now the pod are responsible for reporting the state of their database to the operator, by updating their associated galera CR via calls to the API server. This allows asynchronous reporting from the operator's perspective, and ensure that newly (re)started pod will always update the operator state, preventing stale data to linger in the galera CR status.

Jira: [OSPRH-23075](https://redhat.atlassian.net/browse/OSPRH-23075)

### DIFF
--- a/templates/galera/bin/report_local_galera_state.py
+++ b/templates/galera/bin/report_local_galera_state.py
@@ -1,0 +1,170 @@
+#!/usr/bin/python3
+# -*- python -*-
+
+import argparse
+import json
+import logging
+import os
+import re
+import socket
+import ssl
+import subprocess
+import sys
+from time import sleep
+from urllib.error import HTTPError
+from urllib.request import Request, urlopen
+
+LOG_FORMAT = "%(asctime)s - %(filename)s - %(levelname)s: %(message)s"
+
+
+class APIServer:
+    """REST call to the k8s API server and automatic json (un)marshalling"""
+
+    def __init__(self, namespace: str, token: str, cert: str, timeout: int = 10,
+                 api_server: str = "https://kubernetes.default.svc"):
+        self.api_server = api_server
+        self.namespace = namespace
+        self.token = token
+        self.cert = cert
+        self.timeout = timeout
+
+    def get(self, url: str):
+        context = ssl.create_default_context(cafile=self.cert)
+        req = Request(self.api_server + url)
+        req.add_header("Content-Type", "application/json")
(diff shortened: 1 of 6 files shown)


templates/galera/bin/report_local_galera_state.py
```diff
@@ -0,0 +1,170 @@
+#!/usr/bin/python3
```
zzzeek, line 1: i find it unusual there's no .py extension here?  what's the rationale for that?

dciabrin: I originally implemented it in bash, then moved it to python. And since we have this pattern of bind-mounting files directly which is causing us concerns during minor update if anything changes, I figured there was no point it exposing the language here. So I discarded the extension.

zzzeek: OK, so then it's suggestive of a practice we would in theory have for all of our files is that accurate?

dciabrin: That is something that I would like, but I'm open to discussion. When I rewrote this new script in python, I figured all the other scripts would likely benefit from being rewritten in a higher-lever language eventually, and if so, their filename would need to be updated.

So right now I'm in favor of getting rid of file extension in those injected scripts, because we own them, we may need to change their implementation later, and we can set our own convention anyway, so why not.

Do you have a strong opinion on that, maybe we should keep the extension if you think it makes sense?

zzzeek: i sorta like having the extension :)  but the whole "we changed our approach and new bind mounts are a problem" would be the reason we wouldnt, but what is the problem with new bindmounts ?

dciabrin: I'll re-add the extension then, not a big deal.
The problem with adding changing bind-mounts is that we have to special case our code we every new path, because we have to support rolling update, even from an old FRx version to the latest.
(I added https://github.com/openstack-k8s-operators/mariadb-operator/tree/main/test/ansible to verify that new code doesn't break rolling update, it will become a voting prow job in a subsequent PR.)

dciabrin:
prow env seems to be having issues

lmiccini (approved):

PR #474, opened 2026-06-04
Make chainsaw wait timeout configurable

dciabrin:
Some chainsaw test take longer to execute than the default `oc wait` timeout (30s), so make those timeout explicit and default them in the chainsaw config.

### DIFF
--- a/test/chainsaw/config.yaml
+++ b/test/chainsaw/config.yaml
@@ -7,7 +7,7 @@ spec:
     assert: 180s
     delete: 180s
     cleanup: 180s
-    exec: 120s
+    exec: 180s
   execution:
     failFast: true
     parallel: 1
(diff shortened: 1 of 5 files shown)


test/chainsaw/config.yaml
```diff
@@ -7,9 +7,11 @@ spec:
...
     cleanup: 180s
-    exec: 120s
+    exec: 180s
   execution:
     failFast: true
     parallel: 1
   deletion:
     propagation: Foreground
+  values:
+    ocWaitTimeout: (160s)
```
abays: I think that since this is a literal value, it might need to be `160s` without parens?

dciabrin: Right I pushed before testing, this does not work anyway, so I moved that constants into tests themselves.

dciabrin:
CI test timed out before chainsaw could run

abays (approved):
/lgtm

PR #477, opened 2026-06-08
Improve chainsaw error reporting

dciabrin:
On error, chainsaw deletes all resources that were created during the execution of a run, so it is difficult to pinpoint an error as CI jobs capture must-gathers which are already empty.

Add a script for error reporting so errors happening during a run can be reported as a sort of exception.

Also improve the cleanup steps of currently failing script, so they can easily be run locally with --skip-delete option without blocking.


dciabrin:
Last run passed, trying again to check if it is stable

lmiccini (approved):

PR #485, opened 2026-06-23
Use the latest installer for golangci-lint

dciabrin:
Lately, golangci-lint started providing sbom.json files along with the tarball that we are downloading for running golangci linters, and the version of the installer that we are relying upon fails to validate the tarball hash because of that.

As explained in golangci-lint changelog [1], we should now fetch the installer from a new URL to fix the issue.

[1] https://golangci-lint.run/docs/product/changelog/#v2121

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -121,7 +121,7 @@ tidy: ## Run go mod tidy on every mod file in the repo
 GOLANGCI_LINT_VERSION ?= v2.7.2
 .PHONY: golangci-lint
 golangci-lint:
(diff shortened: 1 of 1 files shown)


dciabrin:
/retest
/go/pkg/mod/golang.org/x/term@v0.40.0/term_unix.go:10:2: golang.org/x/sys@v0.41.0: read "https://proxy.golang.org/golang.org/x/sys/@v/v0.41.0.zip": stream error: stream ID 107; INTERNAL_ERROR; received from peer

stuggi (approved):
/lgtm

PR #492, opened 2026-06-30
Add MariaDB major version upgrade support (OSPRH-31231)

zzzeek:
Adds spec.targetVersion to the Galera CR so the openstack-operator can signal a major version upgrade (e.g. "10.11"). When targetVersion differs from the deployed version recorded in status.clusterProperties, the controller adds TargetVersion to the ClusterProperties hash, triggering a full cluster stop via the existing StopRequired mechanism. On restart, an upgrade init container runs mysql_version_upgrade.sh before mysql_bootstrap.sh, starting mysqld in standalone mode and calling mariadb-upgrade against the node's PVC. Once the cluster bootstraps with the new image, status.clusterProperties["TargetVersion"] is advanced to match spec, removing the init container from subsequent StatefulSet reconciles.

Also adds hack/Containerfile.mariadb-10.11 for building a CentOS 9 based test image with MariaDB 10.11 community packages, to be used for prototyping and validating the upgrade path until a production RHEL 10 image is available.

### DIFF
--- a/hack/Containerfile.mariadb-10.11
+++ b/hack/Containerfile.mariadb-10.11
@@ -1,0 +1,53 @@
+# Test image: MariaDB 10.11 on CentOS Stream 9 kolla base
+#
+# Derives from the standard 10.5 kolla image to preserve all kolla scripts,
+# uid/gid setup, and filesystem layout. Swaps the MariaDB packages to 10.11
+# from mariadb.org community repos.
+#
+# This is a temporary test image for developing/validating the 10.5->10.11
+# upgrade path (OSPRH-31231). The real target image will be RHEL 10 based
+# and built via the production pipeline.
+#
+# Build:
+#   podman build -t quay.io/<user>/openstack-mariadb:10.11-test \
+#     -f hack/Containerfile.mariadb-10.11 .
+#
+# Verify:
+#   podman run --rm quay.io/<user>/openstack-mariadb:10.11-test mysqld --version
+#   podman run --rm quay.io/<user>/openstack-mariadb:10.11-test which mariadb-upgrade
+
+FROM quay.io/podified-antelope-centos9/openstack-mariadb:current-podified
+
+USER root
+
+# Remove the CentOS 9 MariaDB 10.5 RPMs. The community MariaDB-server package
+# conflicts on /usr/sbin/mysqld so these must be removed first.
+RUN rpm -e --nodeps \
+        mariadb-server-galera \
+        mariadb-server \
+        mariadb-server-utils \
+        mariadb-backup \
+        mariadb \
+        mariadb-errmsg \
+        mariadb-common \
+        galera
+
+# Add the MariaDB.org 10.11 repo and install the community packages.
+# MariaDB-server, MariaDB-client, MariaDB-backup use uppercase names so there
+# is no RPM naming conflict with the packages removed above.
+RUN curl -LsS https://downloads.mariadb.com/MariaDB/mariadb_repo_setup | \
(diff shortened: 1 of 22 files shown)


hack/Containerfile.mariadb-10.11
```diff
@@ -0,0 +1,53 @@
...
+        galera-4 && \
+    dnf clean all && \
+    # RHEL packages differ from MariaDB.org community packages in two paths
+    # that the operator scripts and config reference. Add symlinks so this test
+    # image works without touching operator code. The production RHEL 10 image
+    # won't need these.
+    mkdir -p /usr/libexec && ln -s /usr/sbin/mysqld /usr/libexec/mysqld && \
+    mkdir -p /usr/lib64/galera && ln -s /usr/lib64/galera-4/libgalera_smm.so /usr/lib64/galera/libgalera_smm.so
+
+USER mysql
```
zzzeek, line 53: SO important

templates/galera/bin/mysql_root_auth.sh
```diff
@@ -1,6 +1,10 @@
 #!/bin/bash
 set +eu
 
+if [ "$(id -u)" != "$(id -u mysql)" ]; then
```
zzzeek, line 4: this is all stuff I added when i was testing with an image that didn't have "USER: mysql" in it.   i think it's good to have for this kind of thing though.

zzzeek:
I have a public image with mariadb 10.11 on top of RHEL9 at quay.io/rhn_engineering_mbayer/openstack-mariadb:10.11-test-4

templates/galera/bin/mysql_root_auth.sh
dciabrin: ack, it doesn't harm.

internal/controller/galera_controller.go
```diff
@@ -1087,6 +1219,80 @@ func (r *GaleraReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
...
+	//   - Auto-detect: populate the version on clusters that never set
+	//     Spec.TargetVersion, so the status always reflects reality.
+	//   - Upgrade verification: after an upgrade, confirm the running
+	//     version matches what was requested.
+	allReplicasReady := instance.Spec.Replicas != nil &&
+		statefulset.Status.AvailableReplicas == *instance.Spec.Replicas
+
+	var detectedVersion string
+	versionMismatch := false
+	if allReplicasReady && !instance.Status.StopRequired {
```
dciabrin: It took me some time to think about it as this test is ran after the mariadb servers are running.
However it is really only surprising when Spec.Replicas is empty. And in this case, it's fair to run it only once to populate the Status, and it doesn't matter if it's ran after the mysql servers are started and cluster is formed.

dciabrin (approved):
I have tested this one locally by deploying a current upstream based on Centos-9 + mariadb 10.5, and upgrading to UBI-10-based mariadb container image (mariadb 10.11).

I can see all pods stopping at once as expected and everyone running a temporary container mysql-upgrade. Once the upgrade is finished I can see that all pods are reconfigured to remove this temporary container from their definition, which is what is expected.

templates/galera/bin/mysql_root_auth.sh
zzzeek: i had to go searching to see what was here since it was lost.   since the kolla remove rebase, everything runs as mysql hard stop now so we're fine without this stuff.

dciabrin:
cluster could not be claim to run tests

PR #499, opened 2026-07-04
Fix galera bootstrap deadlock when all pods are killed simultaneously

lmiccini:
Three issues combined to delay galera cluster recovery by 12-54+
    minutes when all pods were killed at once.
    
    1. Status update race: injectGcommURI() stored gcomm state in
       Status.Attributes. The deferred PatchInstance (JSON merge patch)
       wrote this back, overwriting pod-pushed attributes (including
       ContainerIDs) with stale data from the start of the reconcile.
    
       Fix: stop writing gcomm state to Status.Attributes entirely.
       Bootstrap-in-progress state is tracked via a sync.Map on the
       reconciler, keyed by the Galera CR's NamespacedName, storing
       only the bootstrap pod name and its containerID at injection time.
       Joiner injection is not tracked at all — the filesystem check in
       isGaleraContainerStartedAndWaiting (test ! -f gcomm_uri && pgrep
       detect_gcomm_and_start.sh) already prevents double-injection,
       since the detect script exec's into mysqld after consuming the file.
    
    2. Unnecessary ContainerID check in findBestCandidate(): the function
       required all replicas to have attributes with ContainerIDs matching
       the currently running containers. But pods restart faster than the
       reconcile cycle, so CIDs never match. This check is unnecessary
       during bootstrap recovery because no pod has started mysqld (pods
       are blocked waiting for gcomm_uri), so the seqno on the persistent
       volume cannot change between container restarts.
    
       Fix: remove the CID comparison from findBestCandidate(). Only
       require that all replicas have pushed attributes with valid seqnos.
       Log CID mismatches for observability without blocking the decision.
    
    3. Spurious joiner push: when all pods were killed, the StatefulSet's
       AvailableReplicas could remain > 0 briefly (stale status), causing
       Bootstrapped to be true. The operator pushed joiner gcomm URIs to
       pods, making them start mysqld against dead peers and wasting a
       restart cycle.
    
       Fix: skip joiner gcomm push when Bootstrapped is set but no pods
       are actually Ready.
    
    Also add a periodic 10s requeue when not bootstrapped, ensuring the
    reconciler retries even when no external events trigger a reconcile.
    
    isBootstrapInProgress returns true only if the container that
    initiated the bootstrap is still present (matching CID). If the pod
    restarted or is gone, the entry is cleared so the operator can
    reprobe. injectGcommURI and getPodsWaitingForGcomm remain free
    functions with no shared state.

Generated-By: claude-opus-4-6

Jira: https://redhat.atlassian.net/browse/OSPRH-32408

### DIFF
--- a/internal/controller/galera_controller.go
+++ b/internal/controller/galera_controller.go
@@ -27,6 +27,7 @@ import (
 	"sort"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/openstack-k8s-operators/lib-common/modules/common"
@@ -91,12 +92,27 @@ var allWatchFields = []string{
 	topologyField,
 }
 
+// bootstrapPodInfo tracks which pod is currently bootstrapping a galera
+// cluster, along with the containerID at injection time so the entry
+// can be invalidated when the container restarts.
+type bootstrapPodInfo struct {
+	podName     string
+	containerID string
+}
+
 // GaleraReconciler reconciles a Galera object
 type GaleraReconciler struct {
 	client.Client
 	Kclient kubernetes.Interface
 	config  *rest.Config
 	Scheme  *runtime.Scheme
+	// bootstrapping tracks the pod currently bootstrapping each Galera
+	// cluster. Keyed by types.NamespacedName. Entries are set when
+	// gcomm:// is injected and cleared when the cluster bootstraps,
+	// the pod restarts (new CID), or the CR is deleted.
+	// Joiner injection is not tracked here — the filesystem check in
+	// isGaleraContainerStartedAndWaiting handles that.
+	bootstrapping sync.Map // types.NamespacedName -> bootstrapPodInfo
 }
 
 // GetLog returns a logger object with a prefix of "controller.name" and additional controller context fields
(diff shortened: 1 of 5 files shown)


internal/controller/galera_controller.go
```diff
@@ -309,26 +356,15 @@ func isGaleraContainerStartedAndWaiting(ctx context.Context, pod *corev1.Pod, in
...
+// the bootstrap state via setBootstrapInProgress after a successful call.
 func injectGcommURI(ctx context.Context, h *helper.Helper, config *rest.Config, instance *mariadbv1.Galera, pod *corev1.Pod, uri string) error {
-	err := mariadb.ExecInPod(ctx, h, config, instance.Namespace, pod.Name, "galera",
+	return mariadb.ExecInPod(ctx, h, config, instance.Namespace, pod.Name, "galera",
 		[]string{"/bin/bash", "-c", "echo '" + uri + "' > /var/lib/mysql/gcomm_uri"},
 		func(_ *bytes.Buffer, _ *bytes.Buffer) error {
-			attr := instance.Status.Attributes[pod.Name]
-			attr.Gcomm = uri
-			instance.Status.Attributes[pod.Name] = attr
 			return nil
```
dciabrin, line 414: The code got inlined too much and got rid of the important comment. Please put it back in.

internal/controller/galera_controller.go
```diff
@@ -959,13 +995,18 @@ func (r *GaleraReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
 	//   . Cluster is bootstrapped as soon as one pod is available
 	instance.Status.Bootstrapped = statefulset.Status.AvailableReplicas > 0
 
+	if instance.Status.Bootstrapped {
+		r.clearBootstrapState(instance)
```
dciabrin, line 1048: Maybe add a comment here to highlight that we clear our transient in-memory bootstrap tracker.

internal/controller/galera_controller.go
```diff
@@ -976,12 +1017,21 @@ func (r *GaleraReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
 		// The remaining pods will become joiner nodes. Wait until they have
 		// reported their status to the operator, and then configure them
 		// as joiners.
+		// Only push joiner gcomm if there are actually Ready pods to
+		// join. If AvailableReplicas is stale and no pods are truly
+		// Ready, pushing a joiner gcomm would cause the pod to start
+		// mysqld against dead peers, wasting a restart cycle.
```
dciabrin: Maybe add "which stales for very long due to the startup probe timeout (240s default)".

internal/controller/galera_controller.go
```diff
@@ -309,26 +356,15 @@ func isGaleraContainerStartedAndWaiting(ctx context.Context, pod *corev1.Pod, in
...
 
-// injectGcommURI configures a pod to start galera with a given URI
+// injectGcommURI configures a pod to start galera with a given URI.
+// For bootstrap URIs (gcomm://), the caller is responsible for recording
+// the bootstrap state via setBootstrapInProgress after a successful call.
 func injectGcommURI(ctx context.Context, h *helper.Helper, config *rest.Config, instance *mariadbv1.Galera, pod *corev1.Pod, uri string) error {
-	err := mariadb.ExecInPod(ctx, h, config, instance.Namespace, pod.Name, "galera",
+	return mariadb.ExecInPod(ctx, h, config, instance.Namespace, pod.Name, "galera",
 		[]string{"/bin/bash", "-c", "echo '" + uri + "' > /var/lib/mysql/gcomm_uri"},
 		func(_ *bytes.Buffer, _ *bytes.Buffer) error {
```
dciabrin, line 413: We no longer record the gcomm uri that we passed to the container to prevent another concurrency update race.
I think we should add a comment here saying we don't need this info because it's been stored directly in the pod, which is the only consumer.

dciabrin (changes requested):
I'm requesting some important comments changes, but the rest seems good to me.
We also might want to add a comment in https://github.com/openstack-k8s-operators/mariadb-operator/blob/main/api/v1beta1/galera_types.go#L130 we don't use it anymore, making type GaleraAttribute a purely read-only attribute from the operator's perspective.

dciabrin (approved):

PR #501, opened 2026-07-08
Bail out during init if the API server is unreachable

dciabrin:
The init container mysql-bootstrap checks whether the root password is still valid and changes it if it required, so that on startup, the galera pod is always able to probe the mysqld process.

This check requires the API server to be up and running. If the API server cannot be reached, no root password check can be attempted, and the init container should error out, to make the pod restart until the API is reacheable again. We do so because the real service container 'galera' requires a password file to be present to run probes, and without the API server the file can't be recreated anyway.

Jira: [OSPRH-29205](https://redhat.atlassian.net/browse/OSPRH-29205)

### DIFF
--- a/templates/galera/bin/mysql_root_auth.sh
+++ b/templates/galera/bin/mysql_root_auth.sh
@@ -83,8 +83,18 @@ GALERA_CR=$(curl -s \
     --header "Authorization: Bearer ${TOKEN}" \
     "${APISERVER}/${MARIADB_API}/namespaces/${NAMESPACE}/galeras/${GALERA_INSTANCE}")
 
+if ! echo "${GALERA_CR}" | python3 -c "import json, sys; json.load(sys.stdin)" 2>/dev/null; then
+    echo "ERROR: failed to retrieve Galera CR from K8s API, cannot determine root password" >&2
+    return 1
+fi
+
 # note jq is not installed in the galera image, macgyvering w/ python instead
 SECRET_NAME=$(echo "${GALERA_CR}" | python3 -c "import json, sys; print(json.load(sys.stdin)['status']['rootDatabaseSecret'])")
+
+if [ -z "${SECRET_NAME}" ]; then
+    echo "ERROR: Galera CR does not contain status.rootDatabaseSecret, cannot determine root password" >&2
+    return 1
+fi
 
 # get password from secret
 PASSWORD=$(curl -s \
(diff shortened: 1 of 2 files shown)


templates/galera/bin/mysql_root_auth.sh
```diff
@@ -83,9 +83,19 @@ GALERA_CR=$(curl -s \
     --header "Authorization: Bearer ${TOKEN}" \
     "${APISERVER}/${MARIADB_API}/namespaces/${NAMESPACE}/galeras/${GALERA_INSTANCE}")
 
+if ! echo "${GALERA_CR}" | python3 -c "import json, sys; json.load(sys.stdin)" 2>/dev/null; then
```
zzzeek, line 86: so this part i get, if we cant get the galera CR at all, no point continuing.  im not sure if you've observed this actually happening, but I get that a pod can spin up and then something goes wrong with the API, sure.

templates/galera/bin/mysql_root_auth.sh
```diff
@@ -83,9 +83,19 @@ GALERA_CR=$(curl -s \
...
+    echo "ERROR: failed to retrieve Galera CR from K8s API, cannot determine root password" >&2
+    return 1
+fi
+
 # note jq is not installed in the galera image, macgyvering w/ python instead
 SECRET_NAME=$(echo "${GALERA_CR}" | python3 -c "import json, sys; print(json.load(sys.stdin)['status']['rootDatabaseSecret'])")
 
+if [ -z "${SECRET_NAME}" ]; then
+    echo "ERROR: Galera CR does not contain status.rootDatabaseSecret, cannot determine root password" >&2
```
zzzeek, line 95: this one im not so sure about, because we have fallthrough logic down below that looks for the secret in the spec if the one in status->rootDatabaseSecret is not valid.   that fallthrough logic is intended for the observed case when a root password change failed, so perhaps the case where "status->rootDatabaseSecret" is just "not there" is separate.  but is this only a theoretical case?

templates/galera/bin/mysql_root_auth.sh
```diff
@@ -95,6 +105,11 @@ PASSWORD=$(curl -s \
     | python3 -c "import json, sys; print(json.load(sys.stdin)['data']['DatabasePassword'])" \
     | base64 -d)
 
+if [ -z "${PASSWORD}" ]; then
+    echo "ERROR: failed to retrieve root password from secret '${SECRET_NAME}', cannot determine root password" >&2
```
zzzeek, line 109: OK this one again is looking for API failure, OK, though it's not distinguishing between the case where the API was not accessible, vs. the case where we got the secret but it didnt have the root PW inside of it.

zzzeek (approved):
I dont quite know what happens if the bootstrap bails out how it comes back and tries again, but if we're going with "something weird happened, so bail out with a nice message and wait for the next [thing zzzeek doesnt understand that will make it try again]", I guess it's fine.    the whole logic I have where it tries Spec->RootDatabaseAccount is for the specific case of "root password update failed", so we have to keep that part for that case separately.

PR #512, opened 2026-07-14
Add PodDisruptionBudget for Galera to prevent quorum loss during rolling reboots

lmiccini:
During an OCP MCP rolling reboot, MCO drains nodes sequentially but does not wait for application pods to become Ready on the previous node before draining the next one. For Galera on a compact (3-node) cluster, this causes quorum loss: the pod on the just-rebooted node is still in Init while the pod on the next node is already being evicted, leaving only 1/3 members running.

Add a PodDisruptionBudget with maxUnavailable=1, matching the pattern used by infra-operator for RabbitMQ. This ensures MCO's drain is blocked until all but one Galera pod are Ready, preventing the overlapping disruption window that causes quorum loss.

The PDB is only created when replicas > 1 and is cleaned up when scaling down to a single replica.

### DIFF
--- a/internal/controller/galera_controller.go
+++ b/internal/controller/galera_controller.go
@@ -36,6 +36,7 @@ import (
 	env "github.com/openstack-k8s-operators/lib-common/modules/common/env"
 	helper "github.com/openstack-k8s-operators/lib-common/modules/common/helper"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/labels"
+	"github.com/openstack-k8s-operators/lib-common/modules/common/pdb"
 	common_rbac "github.com/openstack-k8s-operators/lib-common/modules/common/rbac"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/secret"
 	"github.com/openstack-k8s-operators/lib-common/modules/common/service"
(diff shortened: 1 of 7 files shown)


internal/controller/galera_controller.go
```diff
@@ -713,6 +720,39 @@ func (r *GaleraReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
...
+				condition.ErrorReason,
+				condition.SeverityWarning,
+				condition.PDBReadyErrorMessage, err.Error()))
+			return ctrl.Result{}, err
+		}
+	} else {
+		pdbObj := &policyv1.PodDisruptionBudget{}
+		pdbObj.Name = instance.Name
+		pdbObj.Namespace = instance.Namespace
+		err := r.Delete(ctx, pdbObj)
```
stuggi: use `pdb.DeletePDBWithName(ctx, helper, name, namespace)`  https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/pdb/pdb.go#L158 ?

internal/controller/galera_controller.go
```diff
@@ -713,6 +720,39 @@ func (r *GaleraReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res
...
+	if instance.Spec.Replicas != nil && *instance.Spec.Replicas > 1 {
+		pdbSpec := pdb.MaxUnavailablePodDisruptionBudget(
+			instance.Name,
+			instance.Namespace,
+			intstr.FromInt(1),
+			mariadb.LabelSelectors(instance, "galera"),
+		)
+		pdbInstance := pdb.NewPDB(pdbSpec, 5*time.Second)
+
+		_, err := pdbInstance.CreateOrPatch(ctx, helper)
```
stuggi: if the CreateOrPatch returns NotFound, its not an error https://github.com/openstack-k8s-operators/lib-common/blob/main/modules/common/pdb/pdb.go#L105 ,  it returns the RequeueAfter. Should we return here with this if it happens?

stuggi (approved):
/lgtm

PR #537, opened 2026-08-10
WIP Fix Galera bootstrap with tightened TCIB sudoers

rabi:
Stop using `sudo -E` for `kolla_set_configs` in `mysql_bootstrap.sh`. The TCIB image now relies on explicit `env_keep`, so preserving the full environment is no longer allowed and breaks MariaDB bootstrap.


zzzeek:
we've removed kolla entirely in e2b182e and sudo is no longer used so the change here no longer applies.

